PR #40, opened 2014-11-22
Changes required to use addon-sheepdog

devfaz:
Just .h and .cc files to avoid users being forced to compile opennebula.

### DIFF
--- a/src/vmm/LibVirtDriverKVM.cc
+++ b/src/vmm/LibVirtDriverKVM.cc
@@ -115,6 +115,7 @@ int LibVirtDriver::deployment_description_kvm(
     string  ceph_host       = "";
     string  ceph_secret     = "";
     string  ceph_user       = "";
+    string  sheepdog_host   = "";
     string  gluster_host    = "";
     string  gluster_volume  = "";
 
@@ -434,6 +435,7 @@ int LibVirtDriver::deployment_description_kvm(
         gluster_host    = disk->vector_value("GLUSTER_HOST");
         gluster_volume  = disk->vector_value("GLUSTER_VOLUME");
 
+        sheepdog_host   = disk->vector_value("SHEEPDOG_HOST");
         total_bytes_sec = disk->vector_value("TOTAL_BYTES_SEC");
         read_bytes_sec  = disk->vector_value("READ_BYTES_SEC");
         write_bytes_sec = disk->vector_value("WRITE_BYTES_SEC");
@@ -545,8 +547,7 @@ int LibVirtDriver::deployment_description_kvm(
             {
                 file << "-" << vm->get_oid() << "-" << disk_id;
             }
-
-	    file << "'/>" << endl;
+	    do_network_hosts(file, sheepdog_host, "tcp");
         }
         else if ( type == "GLUSTER" || type == "GLUSTER_CDROM" )
         {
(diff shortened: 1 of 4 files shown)


telmich:
ping - any update on this here? we are very much interested to give opennebula + sheepdog a try

devfaz:
Hi,

Until this is ready in upstream, feel free to use my add-on and/or fork 

https://github.com/OpenNebula/addon-sheepdog

Fabian

rsmontero:
Hi 

I'd like to merge this for the next release. Fabian, in order to make it more general I'm thinking of merging this as is, and rename SHEEPDOG for NETDISK and NETDISK_CDROM. It seems that the KVM code only needs type=network and host.

Make sense? Would you mind to update the addon accordingly? 

If you agree I'll merge this right away

THANKS!!!!!!

devfaz:
Hi,

sure, sounds good.

I will update my code asap.

Fabian

telmich:
Once it is merged, I will give the whole stack a try on our test cluster!

rsmontero:
Fabina sorry! I've just saw the "source protocol='sheepdog'". I think it  is better to keep your patched untouch

jmelis:
In order to make it more generic we could maybe pick up that 'sheepdog'
value from a datastore attribute?

On Thu, Feb 12, 2015 at 12:56 PM, Ruben S. Montero <notifications@github.com

rsmontero:
Yeah, The code for Ceph, Gluster and Ceph is pretty much the same so we need to refactor that part. In any case, this is something that can wait after 4.12

Fabian, there is a new argument for the do_netwotk_hosts function, a default port for the connection , do we have one for Sheepdog? it is for the <host name=... port= > XML element in case the host has not the port part. if there is none, I can just put -1 and require it in the sheepdog_host

rsmontero:
Fabian, changes are now in one-4.10 and master branches with default -1. We can update it with a default.

Could you check that everything is ok?

I'll keep this open till we decide upon the default.

telmich:
Awesome! I will tear down one of our testing infrastructures to test this in the next days!

telmich:
Currently I get

Sun Feb 15 16:18:34 2015 [Z0][VMM][D]: Message received: LOG I 8 error: Failed to open file '/var/lib/one//datastores/0/8/disk.0': No such file or directory

when trying to deploy a VM. The disk.0 is certainly wrong, not sure at which point it should have been translated to sheepdog:...

So far I had to replace "sudo $DOG" with "/usr/sbin/dog" (no sudo, but full path, as oneadmin does not have dog in its path)

telmich:
For some reason I see this in the deployment.0:

                <emulator>/usr/bin/kvm</emulator>
                <disk type='file' device='disk'>
                        <source file='/var/lib/one//datastores/0/8/disk.0'/>
                        <target dev='hda'/>
                        <driver name='qemu' type='qcow2' cache='none'/>
                </disk>

Not yet sure why, because DISK_TYPE == sheepdog should match

src/vmm/LibVirtDriverKVM.cc
```diff
@@ -528,6 +530,25 @@ int LibVirtDriver::deployment_description_kvm(
                      << "\t\t\t</auth>" << endl;
             }
         }
+	else if ( type == "SHEEPDOG" || type == "SHEEPDOG_CDROM" )
+	{
+            if (type == "SHEEPDOG")
```
telmich: In the master tree this reads "sheepdog" (lower case, not UPPER)

telmich:
After changing it to UPPER case and recomping I get

Sun Feb 15 16:43:49 2015 [Z0][TM][D]: Message received: TRANSFER FAILURE 12 Transfer Driver 'sheepdog' not available

PR #42, opened 2014-12-01
OpenNebula vmcontext.sh for FreeBSD guests.

olevole:
Tested on FreeBSD 10.0+

### DIFF
--- a/share/scripts/freebsd/net-vmcontext/vmcontext
+++ b/share/scripts/freebsd/net-vmcontext/vmcontext
@@ -1,0 +1,141 @@
+#!/bin/sh
+# -------------------------------------------------------------------------- #
+# Copyright 2002-2014, OpenNebula Project (OpenNebula.org), C12G Labs        #
+#                                                                            #
+# Licensed under the Apache License, Version 2.0 (the "License"); you may    #
+# not use this file except in compliance with the License. You may obtain    #
+# a copy of the License at                                                   #
+#                                                                            #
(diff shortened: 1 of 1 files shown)


FlorianHeigl:
also found this: https://github.com/slob/freebsd-one-context

FlorianHeigl:
update: works better than what I use now, will try to make the best of both :)

share/scripts/freebsd/net-vmcontext/vmcontext
```diff
@@ -0,0 +1,141 @@
...
+gen_exports() {
+	NETWORK=$1
+	echo "/images -network ${NETWORK}.0/24"
+}
+
+gen_hostname() {
+	MAC=$1
+	NUM=`mac2ip $MAC | cut -d'.' -f4`
+	NUM2=`echo 000000$NUM | sed 's/.*\(..\)/\1/'`
+	sysrc -q hostname="cluster${NUM2}" > /dev/null
```
FlorianHeigl: This doesn't work; unfortunately you can't just assume the system is called "cluster_something"

FlorianHeigl:
it is not fun to autoclose contribs instead of upstreaming them :-( :-(

tinova:
Isn't this covered in the current context addon FreeBSD support?

https://github.com/OpenNebula/addon-context-linux

Let us know so we can close this.

FlorianHeigl:
@olevole can you check please?

tinova:
Please let us know if we should add something to the current FreeBSD support by OpenNebula.

PR #45, opened 2015-01-22
use snapshot if rbd_format is 2 to create consistent images

devfaz:
http://dev.opennebula.org/issues/3521


vholer:
I'm closing this PR, as it can't be applied as is, and reopening back the original issue #872.

PR #54, opened 2015-05-03
Performance improvements for ssh TM_MAD driver

arax:
## Motivation
Accessing and transferring local files via ssh connection causes
significant unnecessary system load (CPU utilization). This patch
takes into consideration data locality and reduces system load by
up to 50 % when transferring large files between the local machine
and a remote machine.

## Patch outline
The local machine is identified by comparing the host part of source
and destination string with the local hostname. In case no match
is detected, old driver behavior is used. If a match is detected
(in source or destination string), the `$HOST:` prefix is removed
and `$PATH` is considered to be locally available. Where ssh-to-ssh
pipeline is used, one part of this pipeline will be executed locally
(not as a remote command through an ssh connection).

## Modified files
 src/tm_mad/ssh/clone
 src/tm_mad/ssh/cpds
 src/tm_mad/ssh/mv
 src/tm_mad/ssh/mvds

### DIFF
--- a/src/tm_mad/ssh/clone
+++ b/src/tm_mad/ssh/clone
@@ -44,11 +44,18 @@ DRIVER_PATH=$(dirname $0)
 # Set dst path and dir
 #-------------------------------------------------------------------------------
 
+WHOAMI=`hostname -f`
+if [ "$?" != "0" ]; then
+    log "Couldn't get local hostname!?"
+    unset WHOAMI
+fi
+
 SRC_PATH=`arg_path $SRC`
 DST_PATH=`arg_path $DST`
 SRC_PATH_SNAP="${SRC_PATH}.snap"
 DST_PATH_SNAP="${DST_PATH}.snap"
 
+SRC_HOST=`arg_host $SRC`
 DST_HOST=`arg_host $DST`
 
 DST_DIR=`dirname $DST_PATH`
@@ -85,7 +92,21 @@ http://*)
     ;;
 
 *)
-    log "Cloning $SRC in $DST_PATH"
+    if [ -n "$WHOAMI" ]; then
+        # check if SRC is the local machine
+        if [ "$WHOAMI" = "$SRC_HOST" ]; then
+            log "Replacing $SRC with $SRC_PATH"
+            SRC=$SRC_PATH
+        fi
+
+        # check if DST is the local machine
+        if [ "$WHOAMI" = "$DST_HOST" ]; then
+            log "Replacing $DST with $DST_PATH"
+            DST=$DST_PATH
+        fi
+    fi
+
+    log "Cloning $SRC into $DST"
     exec_and_log "$SCP -r $SRC $DST" "Error copying $SRC to $DST"
 
     if [ -d "${SRC_PATH_SNAP}" ]; then
--- a/src/tm_mad/ssh/cpds
+++ b/src/tm_mad/ssh/cpds
@@ -38,8 +38,39 @@ fi
 . $TMCOMMON
 
 #-------------------------------------------------------------------------------
+# Set variables
+#-------------------------------------------------------------------------------
+
+WHOAMI=`hostname -f`
+if [ "$?" != "0" ]; then
+    log "Couldn't get local hostname!?"
+    unset WHOAMI
+fi
+
+SRC_PATH=`arg_path $SRC`
+DST_PATH=`arg_path $DST`
+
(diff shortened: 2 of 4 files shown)


arax:
I didn't want to refactor the code too much, to make the changes as transparent as possible. Hence some lines are present in all four files, but we can clean that up quickly if need be.

arax:
Is there something I can improve/change here to help you merge it? Or are you against this change in general?

jmelis:
Hi Boris, we are a bit reluctant to apply the patch as we think it complicates the drivers a bit. The drivers must be simple and efficient, but we all agree that bringing down the number of ssh connections would be an improvement. So we're still evaluating the trade-off in simplicity/performance... but to be honest we are a bit more inclined towards not applying the patch for the moment.

**However** can you tell us (because maybe we are missing something) in what specific actions this patch will be advantageous?

Your example was 'mv', but what about the rest? This is what we can tell:

- mv: agreed, this patch will save one ssh
- clone: we fail to see how this action benefits from the patch. The file starts in the frontend and is scp'd over to the target hypervisor. As far as we can tell the files will never be local (in the same host) and the patch will never have effect.
- cpds: same thing, the files cannot be local
- mvds: same thing

To further clarify our thoughts, let me explain: consider we are running a VM, (and of course feeling quite happy about it :-) ) and then we want to disk-snapshot it (--live) so `cpds` is triggered. `cpds` is a script that is run in the hypervisor, so `whogami` will never be the frontend. We then execute: ``exec_and_log "$SCP -r $SRC $DST" ...`` and since SRC_HOST will **not** be the frontend, and DST_HOST **will** be the frontend, the SSH layer is required.

Am I missing something here?

PS: While I wrote these lines I realised that a possible scenario would be if the frontend was also a hypervisor, but I'm not sure that's the case you have in mind? In any case, that is not a recommended architecture for production and as such we wouldn't like to modify the scripts to cover that possibility

arax:
Hi Jaime,
I'm sorry for responding late :-)

Changes in `clone`, `cpds` and `mvds` will have a similar effect, although only minor. When you use `scp` and both locations are "remote" (local machine specified as `localhost:/path/here/file.here` is still considered to be "remote") then `scp` will open an ssh connection to source and there execute another `scp` for local -> remote transfer. As an example:
```bash
[headnode]: $ scp headnode.example.org:/tmp/test.file workernode.example.org:/tmp/test.file
# will be transformed into
[headnode]: $ scp /tmp/test.file workernode.example.org:/tmp/test.file
# by opening an ssh connection from headnode.example.org to headnode.example.org
# so there are actually two ssh connections active during the whole transfer
```
And that's sort of the point ... I don't want large numbers of pointless headnode -> headnode connections and duplicated `scp` processes. In large deployments, this can cause problems.

There is one significant difference between `mv` and these three ... they are not CPU intensive. Therefore they are not critical and I would be happy to drop these changes in favor of `mv` being accepted :-)

Cheers

arax:
And to address your second question/comment. Yes, frontend/headnode with a hypervisor would be a possible target for this improvement ... but only secondary. This is definitely not designed to improve that ... it's just a side-effect or an extreme use case. This is designed to optimize connection and ssh/scp process handling in large deployments using ssh drivers.

arax:
Did I change your mind or are you still against merging this into master? :-)

jmelis:
> There is one significant difference between mv and these three ... they are not CPU intensive. Therefore they are not critical and I would be happy to drop these changes in favor of mv being accepted :-)

Agreed :) could you redo the patch just for 'mv'? We'd be happy to merge that.

Also, I think you could simplify the PR a bit by dropping entirely the error checking for the hostname:

    WHOAMI=`hostname -f`

    # drop this part...
    if [ "$?" != "0" ]; then
        log "Couldn't get local hostname!?"
        unset WHOAMI
    fi

And substitute this:

    [ -n "$WHOAMI" ] && [ "$WHOAMI" = "$SRC_HOST" ]

with:

    [ "$WHOAMI" = "$SRC_HOST" ]

The second condition implies the first, since "$SRC_HOST" should not be empty, and if it is, that would be the least of our problems :)

I also think that we can do without the `log`s in the code:

    log "$SRC is local, no need to use an ssh connection"
    log "$SRC is remote, using an ssh connection"
    log "$DST is local, no need to use an ssh connection"
    ...

If it fails we will see what command took place, and if it doesn't I don't think it's relevant information in the log files.

Thanks for bearing with us!

PR #55, opened 2015-05-12
rstrip! the endpoint as well

davidk01:
I kept getting bad URI error when loading the endpoint from a file because there was a trailing newline. This fixes that problem.


jfontan:
Merged, thanks!

PR #56, opened 2015-05-19
remotes kvm.rb probe improvement

eucher:
Arch Linux have enp* interfaces.
`free -k` have different output on Arch Linux system, that's why i rewrite memory collecting to /proc/meminfo.
<pre>
free -k
              total        used        free      shared  buff/cache   available
Mem:       32804976    23918256      425892      965480     8460828     7586596
Swap:             0           0           0
</pre>
And in current code calculating cpu usage from `top -bin2` output parsed wrong when system locale set to something else than en_EN. Commas instead of dots when `LANG=ru_RU.UTF-8`:
<pre>
%Cpu(s):  1,3 us,  0,5 sy,  0,0 ni, 98,1 id,  0,0 wa,  0,0 hi,  0,0 si,  0,0 st
</pre>
There are all this values in `/proc/stat`.

### DIFF
--- a/src/im_mad/remotes/kvm-probes.d/kvm.rb
+++ b/src/im_mad/remotes/kvm-probes.d/kvm.rb
@@ -43,7 +43,7 @@ nodeinfo_text.split(/\n/).each{|line|
 #   for everything else, top & proc
 #####
 
-NETINTERFACE = "eth|bond|em|enp2|p[0-9]+p[0-9]+"
+NETINTERFACE = "eth|bond|em|enp|p[0-9]+p[0-9]+"
 
 stat = `cat /proc/stat`
 
--- a/src/im_mad/remotes/kvm-probes.d/kvm.rb
+++ b/src/im_mad/remotes/kvm-probes.d/kvm.rb
@@ -50,7 +50,7 @@ stat = `cat /proc/stat`
 stat.each_line do |line|
   if /\Acpu (.*)\Z/.match(line)
     stat_cpu = Regexp.last_match[1].to_s.split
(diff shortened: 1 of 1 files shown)


jfontan:
I still have to further test the code. It mostly looks good but there's a problem with the CPU information.

`stat` file gives CPU information since the machine was started, that is, average values. For monitoring we need real time values as we want to check for peaks. That's why we used `top` as it measures CPU consumption for 1 second.

I see several ways of fixing this:
* Get CPU information two times separated by a second and check the real consumption.
* Save CPU `stat` values between monitoring cycles in the node plus the timestamp and use them as a delta ( `(value_now - last_value) / (time_now - last_time)` ).
* Use top.

eucher:
Oh yea, since started...
May be last line from `vmstat 1 2` output be right?

jfontan:
Great! I've tested it in Arch, Ubuntu 12.04, CentOS 6 and 7. The only change I would make is this line:

```ruby
nodeinfo_text = `LANG=C virsh -c qemu:///system nodeinfo`
``` 
to:

```ruby
ENV['LANG'] = 'C'

nodeinfo_text = `virsh -c qemu:///system nodeinfo`
```

In ruby 1.8 the command string is not sent to `bash` but directly executed so it didn't work.

Thank you!

jfontan:
It's now in master. Thank you!

hoonetorg:
I've seen this pull request is not yet in 4.12.2. It would be nice, if it could get in there, because CentOS 7.1 needs the memory fix (output of the free command changed). At the moment I manually patch the kvm.rb script, before connecting hypervisors. Without the patch hypervisors have no free mem and no instance could be started.

jfontan:
I've just cherry picked the commits to one-4.12 branch

PR #58, opened 2015-06-11
Feature #3666: Clean VXLAN devices when no VM is left in the hypervisor

goberle:
This patch allows to clean VLAN and VXLAN devices, as well as the bridge when there is no more virtual machines in the bridge.

Any comments or suggestions are welcome. :)


jmelis:
Hi, thanks for the patch, we will merge this shortly.

jmelis:
Merged. Thanks for the great contribution :)

PR #62, opened 2015-06-24
Bug #3846: Already existing rules in security group driver

goberle:
When a NIC is attached, we only try to add security group associated to that NIC.


jmelis:
thanks!

PR #65, opened 2015-07-06
feature-3060: enable disk discard option for qemu-kvm

atodorov-storpool:
This is proposed solution for feature in Backlog #3060 (http://dev.opennebula.org/issues/3060)

By default this feature is not touching the xml configuration for backward compatibility.

There is global VM config in vmm/remotes/kvm/kvmrc named `DEFAULT_ATTACH_DISCARD` that if set will enforce the discard option to all VMs if there is no configuranito per template.

It is possible to enable the discard feature per VM in their template by adding
`DISCARD=unmap`
to enable and
`DISCARD=ignore`
to disable the discard

### DIFF
--- a/share/man/onevm.1
+++ b/share/man/onevm.1
@@ -67,6 +67,8 @@
  \-\-cache cache_mode        Hypervisor cache mode: default, none,
                            writethrough, writeback, directsync or unsafe\.
                            (Only KVM driver)
+ \-\-discard discard_mode    Hypervisor discard mode: ignore or unmap\.
+                           (Only KVM driver)
  \-n, \-\-network id|name     Selects the virtual network
  \-i, \-\-ip ip               IP address for the new NIC
  \-l, \-\-list x,y,z          Selects columns to display with list command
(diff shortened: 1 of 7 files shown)


rsmontero:
Now in master thanks

atodorov-storpool:
a patch slipped out bug in the discard= construction - there is an extra single comma (') slipped out that I fixed on the buld host but obviously skipped to update on the git source.
in the patch for  src/vmm/LibVirtDriverKVM.cc:
    file << "' discard='";
should be:
    file << " discard='";

What is the proper way? add the patch and generate the pull request again?

rsmontero:
I'll fix it

rsmontero:
Here, can you check it?

commit 997eaa029b886fe2dd33608862482353b469de4b
Author: Ruben S. Montero <rsmontero@opennebula.org>
Date:   Wed Jul 8 11:24:58 2015 +0200

    feature #3060: Fix wrong format of discard. Thanks to Anton Todorov

atodorov-storpool:
Yes, this way the cache option is OK without second quotation (') at the end of the parameter.

PR #67, opened 2015-07-06
add LISTEN_ADDRESS config for oned XMLRPC

jtriley:
This allows oned's XMLRPC server to listen on specific interfaces instead of all interfaces. This is useful for binding oned to localhost and using an HTTPS proxy to secure XMLRPC without requiring iptables to restrict unsecured public access. The new `LISTEN_ADDRESS` config defaults to `0.0.0.0` (ie `INADDR_ANY`) for backwards compatibility.

### DIFF
--- a/include/RequestManager.h
+++ b/include/RequestManager.h
@@ -51,6 +51,7 @@ public:
             int _timeout,
             const string _xml_log_file,
             const string call_log_format,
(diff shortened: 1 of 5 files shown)


rsmontero:
Now in master, THANKS!

PR #69, opened 2015-08-19
Fix accountig default date problem

Ara4Sh:
Resolving error : Time range end is not a valid date. It must be YYYY/MM/DD


dmamolina:
This has been already fixed for the next 4.14 version. Thank you

PR #70, opened 2015-08-31
improve rescue of VM when node fail is detected

atodorov-storpool:
* Introduce new LCM states PROLOG_MIGRATE_UNKNOWN and PROLOG_MIGRATE_UNKNOWN_FAILURE

* Change VM migrate logic when state is ACTIVE and lcm_state is UNKNOWN to
  call TM's PROLOG_MIGR action before VMM's BOOT

All core TM drivers that are not empty are skipping disks so there is no impact on
the default behavior

The datastore addon drivers that implement access to raw block devices should check
if the LCM_STATE == 60 (PROLOG_MIGRATE_UNKNOWN) and to remove block device access
from the failed node and provide access to the current node.
There is a simple script function added to get LCM_STATE that can be used as follow

```bash
LCM_STATE=$(lcm_state)
if [ "$LCM_STATE" = "60" ]
then
# remove access from SRC
# add access for DST
fi
```


atodorov-storpool:
This pull request should be linked to [feature request](http://dev.opennebula.org/issues/3958) #3958 but I messed the "Pull request" field there.

rsmontero:
This is no merged in master. THANKS!

PR #71, opened 2015-09-03
Sunstone kerberos support

alvarosimon:
This PR includes a new sunstone auth script called `remote` to support Kerberos/freeip. It's similar to X509 auth, the auth work is still done by apache. Kerberos users should be included using x509 auth using username@DOMAIN instead of certificate DNs.


jfontan:
Cherry picked into one-4.14 and master branches. Thanks!

PR #72, opened 2015-10-26
fs_lvm without CLVM 

apocas:
Let's try to ditch the CLVM requirement from the `fs_lvm` drivers.

`shared` drivers needed to be tweaked, since `fs_lvm` works with two datastores: a system with type `shared` and image one with type `fs_lvm`.

During live migration the `shared` tm driver is called, `shared` loads the necessary data and calls the correct `tm` driver. (basically `shared` is now a dispatcher'ish for other shared `tm` drivers)

Theoretically this shouldn't break drivers different than `fs_lvm`, but it should be tested nevertheless.

`bc` command was used to do some basic math, a bash guru (which I'm not) should be able to drop this requirement. (frontend should feature the `bc` command ex: `yum install bc`)

Only live migrate works for now, offline migration does not work because `pre` and `postmigrate` `tm` scripts are not called during offline migration.
`vmm` could be tweaked to be `lvm` aware but then it would break compatibility with other scenarios/setups which is a no go.


jmelis:
Hi Pedro,

thanks a lot for this PR. We intend to merge it into the master branch and include it into OpenNebula 5.0. We are leaving this open for a while until we can start with this campaign, but I just wanted to let you know that we are definitely interested in including this!

thanks!

jfontan:
I've merged your changes and made changes in branch [feature-4110](/OpenNebula/one/tree/feature-4110). Some of the changes done:

* `fs_lvm` is now a system datastore TM
* Uses `lvscan` instead of `lvs`. In Ubuntu the command `lvs` does create new device files.
* Enabled cold migrations
* `dd` commands don't use `sudo`. Using `sudo` makes the commands fail in nfs shares with `root_squash` and it is not needed as `oneadmin` should be in `disk` group
* Uses `qemu-img` to create new images. The old method (using `dd`, skip N megabytes and writting 1 byte) creates images of N megabytes plus one byte. Logical Volumes must be multiple of 512 bytes so the size must be exact

PR #73, opened 2015-11-04
Bug #4125: Probes cleaned by systemd-tmpfiles

dkelleher:
Add systemd-tmpfiles exemption for /var/tmp/one to prevent tmpfiles from cleaning the probes.  This was tested by reducing the timer on /usr/lib/tmpfiles.d/tmp.conf and adding this exclusion to some nodes but not all.

The idea was to include this in the opennebula-node-kvm package.


jfontan:
I've made a cherry pick to one-4.14 and master branches. Thanks!

PR #74, opened 2015-11-05
bug-4137 enable possibility to predefine variables

atodorov-storpool:
This is proposition for a general solution for solving bug #4137

For example on CentOS 7 to create XFS filesystem on block device set in the environment or befo the script is called:

MKFS="sudo /sbin/mkfs"


jfontan:
Cherry pick to one-4.14 and master branches. Thanks!

PR #75, opened 2015-11-05
When monitoring ceph disks, prefers `rbd ls` to `rbd info`.

pariviere:
`rbs ls` produces lots of useless IOPS. `rbd info` should be prefered
even if it means more rbd calls (results can not be cached as with `rbd ls`)

---

I don't realy know if the `overlap` value if revelant for snapshot size but in my case it matches

### DIFF
--- a/src/vmm_mad/remotes/poll_xen_kvm.rb
+++ b/src/vmm_mad/remotes/poll_xen_kvm.rb
@@ -342,26 +342,24 @@ module KVM
                 pool, image = name.split('/')
                 disk_id = image.split('-')[-1].to_i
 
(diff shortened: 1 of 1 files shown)


a-nldisr:
+1 
Our ceph cluster went from 2-8k ops to 60k ops due to 4.14 release. We applied this patch but it should have been in official release at start.

a-nldisr:
wow... just, wow.

rsmontero:
This has been greatly improved since 4.14 release. Recent versions of Opennebula do not use rbd info, but the overall strategy has changed, so the monitor process is not executed for every single VM from different hypervisors. It is now a single monitor thread for the pool, that uses ceph df to gather the information.

PR #76, opened 2015-11-08
allow block devices for SYSTEM_DS images

atodorov-storpool:
There are block devices that can be used as system datastore like:

CLVM
iSCSI
StorPool

This patch is in need to test and develop our addon against the 'master' branch.
Probably better solution is to have the list of supported devices as list in the configuration files.

### DIFF
--- a/src/datastore/Datastore.cc
+++ b/src/datastore/Datastore.cc
@@ -344,12 +344,12 @@ int Datastore::set_ds_disk_type(string& s_dt, string& error)
             {
                 //Valid disk types for System DS
                 case Image::FILE:
+                case Image::BLOCK:
                 case Image::RBD:
                     break;
 
                 case Image::GLUSTER:
                 case Image::SHEEPDOG:
-                case Image::BLOCK:
                 case Image::CD_ROM:
                 case Image::RBD_CDROM:
                 case Image::SHEEPDOG_CDROM:


rsmontero:
Hi

We have just merge some functionality (branch feature-3987) to deal use Ceph as a System DS. The patch allow to use volatile disks as Ceph volumes. The changes are basically:

- Volatile disks get DS attributes (inherited) so you can get DISK_TYPE in the volatile disk to tune the mkfs actions.
- restore/save actions got TM hooks. So restore.<tm_mad> is call before restoring and same for save. You can use this to put the checkpoint in storpool 
- System DS monitor has been improved. (shared System DS are monitored once from the front-end/BRIDGE_LIST you need a monitor script in the TM directory)
- There is a new attribute ds_migrate to enable/disable system ds migrations.

Could you check against the current HEAD at master and see if the changes suits storpool driver. Any feedback will be more than welcome!

Going back to the patch, we introduced the GLUSTER, RBD, SHEEPDOG to deal with deployment file generation. Currently the block devices are managed through links qemu uses the devices through this links... so I am not sure about this patch.

THANKS 

Ruben

atodorov-storpool:
Hi Ruben,

Thank you for your prompt response.

Somehow I've missed the possibility to get DS attributes. I'll test it and report back. I was concerned is it a repeat of the disk polling case where I am patching vmm/kvm/poll to enable reporting for StorPool.
Anyway I like the SYSTEM_DS support because it will enlighten some dark magic I am doing in our driver to enable context and volatile disks on StorPool for 4.12+ and we have customer using it :)

Cheers
Anton

PR #79, opened 2015-12-29
fix ISO symlink for VMware CDROMs

atodorov-storpool:
While digging another issue, I've spot that when there is SYSTEM_DS migration (at least on the shared TM_MAD) the symlink became broken because it is pointing to the absolute path, which is not available after the migration.

The proposed patch is creating the symlink relative so there is no issue when the absolute path is changed.

The patch is against current master branch but the issue exists in the current stable (4.14.2) too.


rsmontero:
Thanks!

PR #80, opened 2015-12-30
fix ssh_make_path function

atodorov-storpool:
Check against exec rc from the ssh session.


rsmontero:
Thanks, it's clearer after the patch

PR #84, opened 2016-02-16
Bug #4337: IP spoofing filters DHCP communication

vholer:
Don't filter UDP DHCP traffic from 0.0.0.0/32 port 68 to
255.255.255.255/32 port 67.


jmelis:
thanks!

PR #85, opened 2016-02-22
Fixed restart-sched in opennebula init script

arax:
Fixing wrong use of `do_start` in restart-sched routine.


rsmontero:
Thanks!

This is now in master +  one-4.14 branches

PR #86, opened 2016-03-02
Proposal: changing lvremove behavior

apocas:
PROPOSAL:
lvremove -f can lead to race conditions at udev making all the lvm commands to hang indefinitely.

Nowadays lvm defaults to retry_deactivation = 1 at lvm.conf, which lvremove uses and theoretically should mitigate this.
However I recently had issues even with retry_deactivation.

Probably a safer bet is to deactivate the volume before removing it, if the deactivation fails the removal process should fail.
This way even if the -f behavior change in the future we will be OK.

In OpenNebula's lvm use case this shouldn't have side effects, however feedback is needed on this.


rsmontero:
Hi Pedro

Thanks for this !!!! we are going to merge this together with your contribution to get rid of cLVM in pull request #72 

Cheers

ccesario:
Nice PR, this fix my problem related to Instance remove process when using fs_lvm. 
The instance is removed, but the Logical volume remain, then it is needed remove Logical volume manually. With this patch, the Instance and Logical Volumes are removed as expected.

rsmontero:
Seems this is no longer needed

PR #91, opened 2016-04-11
Feature #4399: Add VNC/Spice keymap option to the CLI

goberle:
http://dev.opennebula.org/issues/4399


jfontan:
Thanks! Man pages will be recreated from command help just before release anyway.

PR #100, opened 2016-06-09
im/monitor_ds.sh skip broken symlinks

atodorov-storpool:
Most probably my test setup is too weird but if you have
_/var/lib/one/datastores_ shared on all hosts and want to have 
SYSTEM datastore with TM_MAD on one of the hosts, I am doing the following:

```
mkdir -p /var/lib/one-local/102
ln -s /var/lib/one-local/102 /var/lib/one/datastores/102
```

This way I have "local" datastore nested in "shared" filesystem.

In the above scenario on all other hosts the symlink is dead.

The proposed patch test the is the $dir directory and continue only if it is indeed directory.


jfontan:
Thanks!

PR #101, opened 2016-06-15
Add support for LDAP authentication RFC2307/RFC2307BIS One 4.14

titansmc:
Hi,
I found out that there is only one standard implemented (to be able to do group mapping), the one that every username has an attribute memberOf with the name of the groups that he/she belongs to, but not the other standard that works the other way around, on every group there is an attribute "memberUid" that contains the usernames of the people that belongs to this group.

/opt/opennebula/latest # vi lib/ruby/opennebula/ldap_auth.rb

So, is there a patch I can apply? or can you pull this request to the next release?

Thanks.

### DIFF
--- a/include/Nebula.h
+++ b/include/Nebula.h
@@ -528,36 +528,6 @@ public:
         return -1;
     };
 
-    /**
-     *  Gets a VM_MAD configuration attribute
-     */
-    int get_vm_conf_attribute(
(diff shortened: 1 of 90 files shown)


tinova:
Please open a feature request in the development portal of OpenNebula:
  
   http://dev.opennebula.org/

A pull request is usually to contribute functionality.

PR #102, opened 2016-06-16
Added OpenvSwitch fake bridge support 

RaveNoX:
Added OpenvSwitch fake bridge support.
Fake bridge - bridge which created by ovs-vsctl add-br BRIDGE PARENT VLAN

### DIFF
--- a/src/vnm_mad/remotes/ovswitch/OpenvSwitch.rb
+++ b/src/vnm_mad/remotes/ovswitch/OpenvSwitch.rb
@@ -34,7 +34,18 @@ class OpenvSwitchVLAN < VNMMAD::VNMDriver
             if nic[:bridge_ovs] && !nic[:bridge_ovs].empty?
                 nic[:bridge] = nic[:bridge_ovs]
             end
+            nic[:bridge_ovs_real] = get_real_bridge(nic[:bridge])
         end
+    end
+
+    def get_real_bridge(bridge)
+        ret = nil
+        while ret != bridge do
+            bridge = ret unless ret == nil
+            ret = `#{command(:ovs_vsctl)} br-to-parent #{bridge}`.strip
+        end
+
+        return ret
     end
 
     def activate
@@ -164,7 +175,7 @@ class OpenvSwitchVLAN < VNMMAD::VNMDriver
 
         in_port = ""
 
-        dump_flows = "#{command(:ovs_ofctl)} dump-flows #{@nic[:bridge]}"
+        dump_flows = "#{command(:ovs_ofctl)} dump-flows #{@nic[:bridge_ovs_real]}"
         `#{dump_flows}`.lines do |flow|
             next unless flow.match("#{@nic[:mac]}")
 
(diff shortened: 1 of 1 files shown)


jmelis:
Hi, can you please add a bit of rationale to this PR? Why do you want to have an integration bridge?

thanks!!

RaveNoX:
I have this ovs configuration:
```
# ovs-vsctl show
5650f5d8-ffed-497a-acfa-4aac07b4af5a
    Bridge ovs-br-agg
        Port "vnet2"
            tag: 301
            Interface "vnet2"
        Port ovs-br-int
            tag: 301
            Interface ovs-br-int
                type: internal
        Port "vnet0"
            tag: 301
            Interface "vnet0"
        Port "vnet1"
            tag: 301
            Interface "vnet1"
        Port agg-int
            Interface "p2p1"
            Interface "p1p1"
        Port ovs-br-ceph
            tag: 302
            Interface ovs-br-ceph
                type: internal
        Port "vnet5"
            tag: 301
            Interface "vnet5"
        Port ovs-br-agg
            Interface ovs-br-agg
                type: internal
        Port "vnet7"
            tag: 301
            Interface "vnet7"
        Port "vnet3"
            tag: 301
            Interface "vnet3"
    ovs_version: "2.0.2"
```

As you can see *ovs-br-int* is "fake bridge" which is bridge with tag (vlan) support, if you attach *vnet* to his bridge, it is really will be attached to parent bridge *ovs-br-agg* not to *ovs-br-int*.

As result of this behavior of ovs without my pull request deployment of new instances or migration of already exiting instances will fail with error on adding / removed flows with *ovs-ofctl* because fake bridge is not real bridge and flows which must be set on it must be set on parent real bridge instead of this.

This pull request adding search for parent real bridge for ovs-ofctl commands.

jmelis:
Sorry to bother you a bit more :)

So the BRIDGE in your vnet template is ovs-br-int then, and not ovs-br-agg? You mention this is to add vlan support, however, the OVS already handles vlan support, doesn't it? What extra functionality are you  obtaining by using a fake bridge in the vnet template and not the parent bridge?

(I'm all in favour of merging this PR, but I want to understand it first)

RaveNoX:
Sorry for long wait for an answer.

>So the BRIDGE in your vnet template is ovs-br-int then, and not ovs-br-agg
yes

>You mention this is to add vlan support, however, the OVS already handles vlan support, doesn't it?
Yes you are right. 
Before current configuration we are first use configuration where is no aggregation was used.
ovs-br-agg and ovs-br-ceph was owners of physical interfaces each, and bridges configuration keeping from it.
I don't know any way for changing nodes network configuration without affecting running instances.

Maybe this PR may be also useful for some custom ovs configuration for certain situations (maybe for STP or other settings which must be other than on physical-bounded ovs bridge).

Anyway we are already using this code in our production systems which are running under One 5.0 and 4.14 (I also have backport of this PR for it).
It will be great if for future releases this code will be already included in ONE.

PR #104, opened 2016-06-22
fix for bug #4593

atodorov-storpool:
Hi Developers,
http://dev.opennebula.org/issues/4593
When VM in UNKNOWN state is rescheduled it can not be migrated anymore because a check in VirtualMachineMigrate::request_execute is not satisfied due to previous history record not closed properly.
https://github.com/OpenNebula/one/blob/master/src/rm/RequestManagerVirtualMachine.cc#L1097

The suggested patch updates(closes) previous history record properly (I hope) and the VM is in proper state then.

Kind Regards,
Anton Todorov

### DIFF
--- a/src/lcm/LifeCycleStates.cc
+++ b/src/lcm/LifeCycleStates.cc
@@ -682,10 +682,14 @@ void LifeCycleManager::prolog_success_action(int vid)
                     vm->set_state(VirtualMachine::BOOT_MIGRATE);
                     break;
 
-                case VirtualMachine::PROLOG:
-                case VirtualMachine::PROLOG_FAILURE: //recover success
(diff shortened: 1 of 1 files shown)


rsmontero:
Needs to handle the other PROLOG_MIGRATE states and failure.

rsmontero:
Hi Anton,

After reviewing this, I've decided to move your changes to the action. This case is slightly different to the others as the capacity is decremented before issuing the prolog. So I prefer to treat the history record in a similar way to a POWEROFF/SUSPEND migration, where the actual registry is closed. All changes are now in master if you want to give them a try, but I basically moved the update from the callback to the trigger.

THANKS!

PR #105, opened 2016-07-07
Update clone

lejmr:
For the case when system datastore (0) is not LVM based then I get erros.. According to the documentation I believe this is the proper coding


jfontan:
`fs_lvm` driver must use an LVM system datastore. Can you elaborate in your use case?

PR #106, opened 2016-07-14
fix for FT action when host is in ERROR state

atodorov-storpool:
Following https://forum.opennebula.org/t/vm-fails-to-ha-reschedule-on-host-error-opennebula-5-0-1/2566

Skip access to the source host when called in LCM_STATE=PROLOG_MIGRATE_UNKNOWN because the source host is not reachable.

It is used in CEPH's TM_MAD but I think it is OK for SSH too.

### DIFF
--- a/src/tm_mad/ssh/mv
+++ b/src/tm_mad/ssh/mv
@@ -71,6 +71,11 @@ log "Moving $SRC to $DST"
 ssh_exec_and_log "$DST_HOST" "rm -rf '$DST_PATH'" \
     "Error removing target path to prevent overwrite errors"
 
+if [ `lcm_state` -eq 60 ]; then
+    log "Not moving files from $SRC_HOST in FT mode"
+    exit 0
+fi
+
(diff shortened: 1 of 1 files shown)


jmelis:
Thanks for this PR! we are going to wait for 5.2 to merge this and improve the ft hook in general:
http://dev.opennebula.org/issues/4659

jmelis:
thanks! merged

HenriWahl:
Seems to be missing in tm/ceph/mv - shouldn't it be there too according to https://forum.opennebula.org/t/vm-fails-to-ha-reschedule-on-host-error-opennebula-5-0-1/2566 ?

atodorov-storpool:
In the sources src/tm_mad/ceph/mv is symlink to tm_mad/ssh/mv and should be expanded when packaged.

@jmelis

HenriWahl:
Sorry I was looking in the wrong lines...

PR #107, opened 2016-07-15
vmm/kvm/poll cleanup

atodorov-storpool:
Cleaned up unused code.
The rbd stats monitoring are moved to monitor script.


jfontan:
That's so true. Thanks!

PR #108, opened 2016-08-26
Bug #4748: Fix sunstone remote_auth login

alvarosimon:
This is a small fix to solve sunstone remote_user login issue.

http://dev.opennebula.org/issues/4748


src/sunstone/public/app/opennebula/auth.js
```diff
@@ -39,7 +39,9 @@ define(function(require) {
           var authString = 'Basic ';
 
           authString += btoa(unescape(encodeURIComponent(token)));
-          req.setRequestHeader("Authorization", authString);
+          if (username && password) {
```
stdweird: you can move this conditional to the beginning of the function?

PR #109, opened 2016-08-31
fix gem install for thin on ruby < 2.2.0

hoonetorg:
issue when trying on el7 with ruby 2.0


hoonetorg:
Oh, I don't understand what the  problem is here. Isn't this already merged?

tinova:
Hi! This hasn't come up in our certification against currently supported platforms, which in include RHEL7. Could you elaborate a bit more on the error you are experiencing?

PR #110, opened 2016-09-02
MAD scripts must log command outputs

baby-gnu:
Actually the “EXEC_LOG_ERR” variable is always empty since we redirect
stderr to stdout and then redirect stdout to /dev/null.

* src/mad/ruby/scripts_common.rb (OpenNebula.exec_and_log): Do not
  redirect stdout to /dev/null.

* src/mad/sh/scripts_common.sh (exec_and_log): Ditoo.
  (multiline_exec_and_log): Ditoo.
  (exec_and_set_error): Ditoo.
  (ssh_exec_and_log): Ditoo.
  (ssh_exec_and_log_stdin): Ditoo.
  (ssh_make_path): Ditoo.

### DIFF
--- a/src/mad/ruby/scripts_common.rb
+++ b/src/mad/ruby/scripts_common.rb
@@ -59,7 +59,7 @@ module OpenNebula
     # If a second parameter is present it is used as the error message when
     # the command fails
     def self.exec_and_log(command, message=nil)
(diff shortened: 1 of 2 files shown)


jmelis:
Closing as per discussion in http://dev.opennebula.org/issues/4761

PR #112, opened 2016-09-12
Feature #3510: authorize “EMULATOR” in VM TEMPLATE

baby-gnu:
* src/vm/VirtualMachine.cc (insert): Add “EMULATOR” to the list of
  VM attributes understood by opennebula.

* src/vmm/LibVirtDriverKVM.cc (deployment_description_kvm): Get
  “EMULATOR” from VM template first, then from default and finally from
  hardcoded path.

Thanks Vladislav Gorbunov for the patch

### DIFF
--- a/src/vm/VirtualMachine.cc
+++ b/src/vm/VirtualMachine.cc
@@ -472,6 +472,18 @@ int VirtualMachine::insert(SqlDB * db, string& error_str)
     }
 
     // ------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


rsmontero:
Merged thanks! Maybe we can expose this in Sunstone

baby-gnu:
> Merged thanks! Maybe we can expose this in Sunstone

In which case we should manage the ARCH accordingly.

PR #113, opened 2016-09-17
Update failmigrate

atodorov-storpool:
It is obvious that I've totally missed the tm/failmigrate script when proposed the migrate_other function

Here are three commits:
- tm/fs_lvm - besides the call of migrate_other() I think it is good idea to deactivate the LVM volumes enabled by the  `tm/fs_lvm/premigrate` on DST_HOST
- tm/ceph/failmigrate - same as for the above commit - I believe it is good to clean up the DST_HOST if the live-migration fails
- tm/{qcow2,shared,ssh} - do migrate_other() when failmigrate

Kind Regards,
Anton Todorov

### DIFF
--- a/src/tm_mad/ceph/failmigrate
+++ b/src/tm_mad/ceph/failmigrate
@@ -55,6 +55,27 @@ fi
 
 migrate_other "$@"
 
+#--------------------------------------------------------------------------------
+
+# Get custom datastore values
+DRIVER_PATH=$(dirname "$0")
+XPATH="${DRIVER_PATH}/../../datastore/xpath.rb --stdin"
+
+unset XPATH_ELEMENTS i j
+
+while IFS= read -r -d '' element; do
+  XPATH_ELEMENTS[i++]="$element"
+done < <(onedatastore show -x "$DSID" | $XPATH \
+  /DATASTORE/TEMPLATE/SHARED)
+
+SHARED="${XPATH_ELEMENTS[j++]}"
+
+if [ "$SHARED" = "YES" ]; then
+    exit 0
+fi
+
+#--------------------------------------------------------------------------------
+
 ssh_exec_and_log "$DST_HOST" "rm -rf '$DST_PATH'" \
     "Error removing target path to prevent overwrite errors"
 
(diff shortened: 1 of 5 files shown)


atodorov-storpool:
Well IMO the issue is not completely resolved. IMO each core TM_MAD should have algo to call cleanup for other TM_MADs. Otherwise artefacts will be left on the (failed) destination.

rsmontero:
@atodorov-storpool  Trying to understand the migrate_other, it is used to have a list of TM's in the DISK TM_MAD to support disks from different datastore types, but this is not supported. So I'm wondering if this and the migrate other can be remove, or if storpool driver is using this?

atodorov-storpool:
Hi @rsmontero

> @atodorov-storpool Trying to understand the migrate_other, it is used to have a list of TM's in the DISK TM_MAD to support disks from different datastore types, but this is not supported. So I'm wondering if this and the migrate other can be remove, or if storpool driver is using this?

Can't get the _unsupported_ part because this is how most[*] of the addon storage drivers are working with OpenNebula :-)
If I manage to drive you through an example case will help to understand the point. Let's have an OpenNebula installation with SYSTEM datastore TM_MAD _qcow2_ and an additional IMAGE datastore with TM_MAD _storpool_. And a VM could have a disk from the StorPool-backed datastore... When Instantiated on a host, the tm/storpool/ln (or _clone_ if the image is non-persistent) is called to provide the symlink in the VM's home folder (on the qcow2-backed SYSTEM datastore) to the block device. So, on VM migrate the tm/qcow2/premigrate script is called. This script natively does not know how to deal with the storpool disk - here the _migrate_other_ function helps by walking the VM disks and calling the tm/DISK[TM_MAD]/premigrate script that provides the storpool disk on the destination host. Same for postmigrate and failmigrate.

[*] AFAIK [addon-linstor-un](https://github.com/OpenNebula/addon-linstor_un/) maintained by @kvaps and [addon-3par](https://gitlab.feldhost.cz/feldhost-public/one-addon-3par/-/tree/master) maintained by @feldsam are benefiting from this functionality too.

p.s. There was a discussion and a fix for the SSH-backed System datastores to exclude the block devices from the blockcopy, but I can't find it at the moment.

PR #125, opened 2016-10-12
tm/shared/clone cleanup artefacts before copy

atodorov-storpool:
Following https://forum.opennebula.org/t/opennebula-5-0-2-attach-detach-images-datastore-folder-not-cleaned-up/3111

I count this fix as safeguard. IMO the _mvds_ script needs to be extended to cleanup the symlinks when persistent volume is detached.


atodorov-storpool:
already implemented via commit https://github.com/OpenNebula/one/commit/4f9a86bf235c9e55f82f82efee832b069634063e

PR #126, opened 2016-10-12
shared/mvds rm the symlink to the persistent img

atodorov-storpool:
Again, following https://forum.opennebula.org/t/opennebula-5-0-2-attach-detach-images-datastore-folder-not-cleaned-up/3111/11?u=atodorov_storpool

I think that the patch to tm/shared/clone is safeguard and this is proper fix - cleanup the symlink to the persistent image here.

Not tested, but should work. Used ssh/mvds as donor.

### DIFF
--- a/src/tm_mad/shared/mvds
+++ b/src/tm_mad/shared/mvds
@@ -16,4 +16,36 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
+# mvds host:remote_system_ds/disk.i fe:SOURCE vmid dsid
+#   - fe is the front-end hostname
+#   - SOURCE is the path of the disk image in the form DS_BASE_PATH/disk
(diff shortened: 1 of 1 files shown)


atodorov-storpool:
It is qcow2 mvds that should be fixed

atodorov-storpool:
Too early in the morning to close it - tm_mad/qcow2/mvds is symlink to tm_mad/shared/mvds so we should cleanup the symlinks in both places.

rsmontero:
Hi Anton,

After discussing with the team we prefer to keep mvds as an exit 0, and remove the target files before doing any destructive operation. The rationale behind this is to recover from failures while executing MVDS. I've also split the rm -rf $DST_PATH{,.snap} command to not use -r for $DST_PATH, just in case.

Thanks

PR #129, opened 2016-10-18
Update vcenter mkfs - fix #4872

ccesario:
Proposed fix to http://dev.opennebula.org/issues/4872#

Using Sunstone or CLI is not mandatory choose the **disk type** or **adpter type**. If not choose these params, this error happen.

> Tue Oct 18 17:07:34 2016 [Z0][ImM][I]: Not enough information to create the image.
> Tue Oct 18 17:07:34 2016 [Z0][ImM][I]: ExitCode: 255
> Tue Oct 18 17:07:34 2016 [Z0][ImM][E]: Error creating datablock


rsmontero:
This is going to be fix [as part of another issue](http://dev.opennebula.org/issues/4699), and changes integrated from there.

PR #131, opened 2016-10-21
vCenter :: Store empty block images in directory

ccesario:
Proposed fix to add option to create empty block images on specified directory instead root directory

It is needed add FOLDER_IMG attribute on datastore.
ds_option
data_store
image_attached
images1


tinova:
This will be treated as part of #2762

PR #132, opened 2016-10-26
regex match from the beginning to the end of the string in nic_deacti…

winem:
This fixes issue 4882. Actually, we just avoid that interface with id 10  matches lines.match(/^#{chain}/ for example.

### DIFF
--- a/src/vnm_mad/remotes/lib/security_groups_iptables.rb
+++ b/src/vnm_mad/remotes/lib/security_groups_iptables.rb
@@ -328,7 +328,7 @@ module SGIPTables
 
         remove_chains = []
         iptables_s.lines.each do |line|
-            if line.match(/^-N #{chain}/)
+            if line.match(/^-N #{chain}$/)
                  remove_chains << line.split[1]
             end
         end
@@ -336,7 +336,7 @@ module SGIPTables
         remove_chains.each {|c| commands.add :iptables, "-X #{c}" }
 
         ipset_list.lines.each do |line|
-            if line.match(/^#{chain}/)
+            if line.match(/^#{chain}$/)
                 set = line.strip
                 commands.add :ipset, "destroy #{set}"
             end
--- a/src/vnm_mad/remotes/lib/security_groups_iptables.rb
+++ b/src/vnm_mad/remotes/lib/security_groups_iptables.rb
@@ -328,7 +328,7 @@ module SGIPTables
 
         remove_chains = []
         iptables_s.lines.each do |line|
-            if line.match(/^-N #{chain}$/)
+            if line.match(/^-N #{chain}(-|$)/)
                  remove_chains << line.split[1]
             end
         end
@@ -336,7 +336,7 @@ module SGIPTables
         remove_chains.each {|c| commands.add :iptables, "-X #{c}" }
 
         ipset_list.lines.each do |line|
-            if line.match(/^#{chain}$/)
+            if line.match(/^#{chain}(-|$)/)
                 set = line.strip
                 commands.add :ipset, "destroy #{set}"
             end


FlorianHeigl:
@jfontan I heard about this issue on IRC. OK if I merge? patch looks sane to me.

jfontan:
I may be wrong as I have not written that code but the change leaves the chains undeleted. Here's how the chain names are created:

```
        vars[:chain]     = "one-#{vm_id}-#{nic_id}",
        vars[:chain_in]  = "#{vars[:chain]}-i",
        vars[:chain_out] = "#{vars[:chain]}-o"

        if sg_id
            vars[:set_sg_in]  = "#{vars[:chain]}-#{sg_id}-i"
            vars[:set_sg_out] = "#{vars[:chain]}-#{sg_id}-o"
        end
```

And here an example of VM chains (`iptables -S`):

```
-N one-6345-0-i
-N one-6345-0-o
-N one-6345-1-i
-N one-6345-1-o
-N one-6346-0-i
-N one-6346-0-o
-N one-6346-1-i
-N one-6346-1-o
```

The regexp should be something like `/^-N #{chain}-/`. Does it make sense?

I'll try  to reach the people that know more about this part.

winem:
Hi,

 

you are right. I double-checked it and we thought that I have fixed it because the error no longer appeared in the log file. But the chains remain on the system and only the rules will be deleted. 

 

How to proceed here? Successfully tested it with your solution. Should I update my code and push it to my forked repository? Do you want to fork it yourself and provide the update?


Best regards,

Marcel 

 

Gesendet: Donnerstag, 27. Oktober 2016 um 17:23 Uhr
Von: "Javi Fontan" <notifications@github.com>
An: OpenNebula/one <one@noreply.github.com>
Cc: Marcel <mweinberg@email.de>, Author <author@noreply.github.com>
Betreff: Re: [OpenNebula/one] regex match from the beginning to the end of the string in nic_deacti… (#132)


I may be wrong as I have not written that code but the change leaves the chains undeleted. Here's how the chain names are created:

        vars[:chain]     = "one-#{vm_id}-#{nic_id}",
        vars[:chain_in]  = "#{vars[:chain]}-i",
        vars[:chain_out] = "#{vars[:chain]}-o"

        if sg_id
            vars[:set_sg_in]  = "#{vars[:chain]}-#{sg_id}-i"
            vars[:set_sg_out] = "#{vars[:chain]}-#{sg_id}-o"
        end


And here an example of VM chains (iptables -S):

-N one-6345-0-i
-N one-6345-0-o
-N one-6345-1-i
-N one-6345-1-o
-N one-6346-0-i
-N one-6346-0-o
-N one-6346-1-i
-N one-6346-1-o


The regexp should be something like /^-N #{chain}-/. Does it make sense?

I'll try to reach the people that know more about this part.

jfontan:
I'm assigning it to @jmelis as he has better understanding than me about this problem.

jfontan:
The regexp should be `/^-N #{chain}(-|$)/`. Can you change it in your branch?

winem:
Sure. But why is the |$ required? All chains ends with -i or -o. $ at the end of the regex would be equal to my first attempt and no chains would match here. 

I will do the update this evening and give you a ping when it's done.

jfontan:
I was waiting for the response of colleagues to check if there could be a chain without suffixes (that's where `$` would work). To speed things up that regexp would work for both (with and without prefixes).

winem:
Done. 

329         remove_chains = []
330         iptables_s.lines.each do |line|
331             if line.match(/^-N #{chain}(-|$)/)
332                  remove_chains << line.split[1]
333             end
334         end
335         remove_chains.each {|c| commands.add :iptables, "-F #{c}" }
336         remove_chains.each {|c| commands.add :iptables, "-X #{c}" }
337 
338         ipset_list.lines.each do |line|
339             if line.match(/^#{chain}(-|$)/)
340                 set = line.strip
341                 commands.add :ipset, "destroy #{set}"
342             end
343         end

jfontan:
Thanks!

PR #134, opened 2016-10-31
B #4838 change color VNC

juanmont:
- Ctrl-alt-del change to red
 - close change to grey


juanmont:
Choose only the commit new tag and change color VNC

PR #142, opened 2016-11-09
Added support for specifying that the VNC port should be shared

rkeene:
Added support for specifying that the VNC port should be shared the webserver.

This requires some cooperation from the webserver to proxy websocket requests for the VNC/SPICE data to the OpenNebula websocket proxy.


tinova:
worth taking a look, reopening

tinova:
does not apply to FireEdge, closing.

PR #147, opened 2016-11-10
F 4159: IPv6 Security Groups

rkeene:
Basic support for IPv6 Security Groups (untested)


rsmontero:
This is now merged in master. Additionally to this commit, the drivers now make use of IPAddr module getting rid of the IPv4 and IPv6 modules. INET6 family checks are done using IPAddr methods. Also, ipsets are created only if needed for network and network_portrange rules (a rule can only apply either to IPv4 or IPv6 network). Initialization of spoofing IPv4 and IPv6 rules are dealt in the same way.

Thanks

rsmontero:
BTW, I've performed some basic tests to check that the sets and rules are properly installed, some integration tests are still needed as well as Sunstone support

PR #160, opened 2016-11-21
tm/ceph {pre,post}migrate do nothing if not SYSTEM DS

atodorov-storpool:
The proposed solution partially addresses bug #4924

IMO in case of SYSTEM DS on shared filesystem the TM_MAD must be 'shared'.

In this case the tm/ceph/mv will not be called. But then {pre,post}migrate
will be called by migrate-other. In that case there is no work for Ceph tm_mad.


atodorov-storpool:
Tomorrow I'll draft an update for the docs regarding these TM_MAD to describe the "7-th argumet" - the case when the {pre,post}migrate are called as slaves by the primary tm_mad driver.

jmelis:
thanks!!

PR #168, opened 2016-12-02
B#4897 Fix situations that would cause NICs not being removed (one 5.2)

n40lab:
This fix also contains code for fix the duplicated NICs issue that was added to master.


n40lab:
Added a safety check for performance manager interval if it could be negative somehow.

PR #176, opened 2016-12-18
Add support for IPv6 to LISTEN_ADDRESS directive for OpenNebula daemon

rkeene:
The OpenNebula Daemon (oned) currently only ever attempts to listen on IPv4.

This patch adds support for listening on IPv6 if an IPv6-like address is supplied.

### DIFF
--- a/src/rm/RequestManager.cc
+++ b/src/rm/RequestManager.cc
@@ -175,10 +175,40 @@ int RequestManager::setup_socket()
     int socket_af;
     size_t rm_addr_size;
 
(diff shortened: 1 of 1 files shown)


rsmontero:
I've added the IPv6 support through getaddrinfo, as a bonus it now support DNS resolves.

https://github.com/OpenNebula/one/commit/44290311171bf6823319b3c867ab0e72a60279c1#diff-85521f07a5d817d404e1f2567819deb3

PR #182, opened 2017-01-09
Bug #4969: Ceph driver for system datastore ignores SHARED=YES

kvaps:
https://dev.opennebula.org/issues/4969

If you use ceph as system datastore and if your /var/lib/one is shared between your nodes.
Then `premigrate`, `postmigrate` and `mv` scripts from ceph tm driver removes vm folder: /var/lib/one/datastores/{datastoreid}/{vmid} with files like `deployment.0`, `disk.1`, `checkpoint`
This entails the error when `save` vm for `suspend` or `migrate`.
```
Command execution fail: cat << EOT | /var/tmp/one/vmm/kvm/migrate 'one-48' 'c13n1' 'c15n1' 48 c15n1
migrate: Command "virsh --connect qemu:///system migrate --live one-48 qemu+ssh://c13n1/system" failed: error: Cannot access storage file '/var/lib/one//datastores/100/48/disk.1' (as uid:9869, gid:9869): No such file or directory
Could not migrate one-48 to c13n1
ExitCode: 1
```

### DIFF
--- a/src/tm_mad/ceph/postmigrate
+++ b/src/tm_mad/ceph/postmigrate
@@ -61,6 +61,12 @@ source $TMCOMMON
 
 #--------------------------------------------------------------------------------
 
+if [ "$SHARED" = "YES" ]; then
+    exit 0
+fi
+
+#--------------------------------------------------------------------------------
+
 if [ "$SRC_HOST" == "$DST_HOST" ]; then
     log "Not moving $SRC_HOST to $DST_HOST, they are the same host"
     exit 0
@@ -70,8 +76,6 @@ fi
 
 migrate_other "$@"
 
-if [ "$SHARED" != "YES" ]; then
-    exec_and_log "$SSH $SRC_HOST rm -rf $DST_PATH"
-fi
+exec_and_log "$SSH $SRC_HOST rm -rf $DST_PATH"
 
 exit 0
--- a/src/tm_mad/ceph/postmigrate
+++ b/src/tm_mad/ceph/postmigrate
@@ -36,6 +36,21 @@ TEMPLATE_64=$6
 
 #--------------------------------------------------------------------------------
 
+# Get custom datastore values
+DRIVER_PATH=$(dirname "$0")
+XPATH="${DRIVER_PATH}/../../datastore/xpath.rb --stdin"
+
+unset XPATH_ELEMENTS i j
+
+while IFS= read -r -d '' element; do
+  XPATH_ELEMENTS[i++]="$element"
+done < <(onedatastore show -x "$DSID" | $XPATH \
+  /DATASTORE/TEMPLATE/SHARED)
+
+SHARED="${XPATH_ELEMENTS[j++]:-$SHARED}"
+
+#--------------------------------------------------------------------------------
+
 if [ -z "${ONE_LOCATION}" ]; then
     TMCOMMON=/var/lib/one/remotes/tm/tm_common.sh
 else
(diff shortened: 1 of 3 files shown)


src/tm_mad/ceph/postmigrate
```diff
@@ -36,6 +36,21 @@ TEMPLATE_64=$6
...
+XPATH="${DRIVER_PATH}/../../datastore/xpath.rb --stdin"
+
+unset XPATH_ELEMENTS i j
+
+while IFS= read -r -d '' element; do
+  XPATH_ELEMENTS[i++]="$element"
+done < <(onedatastore show -x "$DSID" | $XPATH \
+  /DATASTORE/TEMPLATE/SHARED)
+
+SHARED="${XPATH_ELEMENTS[j++]:-$SHARED}"
```
atodorov-storpool: Excuse me if I am wrong but the SHARED attribute in the datastore template is set via the TM_MAD_CONF definition for the datastore in /etc/oned.conf so it is hard to have missing variable in the template. In addition the bash substitution for `:-$SHARED` expect to have `SHARED` in the environment which I think is not defined in ONE.

src/tm_mad/ceph/postmigrate
```diff
@@ -55,6 +70,8 @@ fi
 
 migrate_other "$@"
 
-exec_and_log "$SSH $SRC_HOST rm -rf $DST_PATH"
+if [ "$SHARED" != "YES" ]; then
```
atodorov-storpool: I think that it is much simpler to have something like

```bash
if [ "$SHARED" = "YES" ]; then
    exit 0
fi
```
This way there are only lines added without change of the indention of the other code.

src/tm_mad/ceph/postmigrate
kvaps: 1) Perhaps I not so understood purpose of this variable, so it is possible to rename to other name, but I consider that this option is necessary for such cases.
Any ideas about it?
2) Fixed

src/tm_mad/ceph/postmigrate
kvaps: OK. I can change it.
But won't it break the code style and create no more confusion there?

atodorov-storpool: I think it is OK because there are other places with similar code style. Take for example

https://github.com/OpenNebula/one/blob/master/src/tm_mad/ssh/mv#L62

src/tm_mad/ceph/postmigrate
atodorov-storpool: Well I've learned it the hard way - I wanted to have single datastore TM_MAD for images and system, but only the images to be shared. So I've tried to edit the SHARED=no in system's record and it was (magically) reverted to SHARED=YES as it was set-up in the /etc/one/oned.conf . Ithink it should be override-able per datastore but I've no time so I made a workaround in our addon.

src/tm_mad/ceph/postmigrate
kvaps: fixed, thanks!

rsmontero:
It seems the issue is oned.conf being to picky about YES/NO in shared. So I think this can be a worksforme

rsmontero:
BTW, please feel free to add anything if you think it is worth reopening it.

rsmontero:
Sorry I misread the issue. Reopening it, to double check if it applies

kvaps:
It seems `SHARED=yes` is not acceptable option for handle this behavior, I've created topic for decide that: 
https://forum.opennebula.org/t/shared-mode-for-the-ceph-based-system-datastires/7164

vholer:
Looks to me this is working, closing.

PR #184, opened 2017-01-13
Nokogiri fails to install for Ruby 2.0.0

n40lab:
When compiling OpenNebula from source code in CentOS 7, /usr/share/one/install_gems fails installing nokogiri.

```
Gem::InstallError: nokogiri requires Ruby version >= 2.1.0.
An error occurred while installing nokogiri (1.7.0.1), and Bundler cannot continue.
Make sure that `gem install nokogiri -v '1.7.0.1'` succeeds before bundling.
```

CentOS 7 provides Ruby 2.0.0p648 which makes the script to install latest nokogiri (1.7.0.1). Nokogiri 1.7.0 requires Ruby >= 2.1 while latest Nokogiri in 1.6.X version (1.6.8.1) requires Ruby >= 1.9.2 so I propose using nokogiri < 1.7.0 if Ruby version is higher than 1.9.2. This allows the install_gems script to finish succesfully.

### DIFF
--- a/share/install_gems/Gemfile
+++ b/share/install_gems/Gemfile
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 
-if RUBY_VERSION < '1.9.0'
+if RUBY_VERSION < '1.9.2'
     gem 'nokogiri', '< 1.6.0'
     gem 'net-ldap', '< 0.9'
     gem 'zendesk_api', '< 1.5'
 else
-    gem 'nokogiri'
+    gem 'nokogiri', '< 1.7.0'
     gem 'net-ldap', '< 0.13'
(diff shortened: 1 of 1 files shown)


jfontan:
This is not documented, maybe this should be added to building from source doc or `install_gems` tool itself.

The way we fix versions in the packages is with `Gemfile.lock` files. There are one per distro in the source. Can you try copying the one for CentOS to the same place where `install_gems` resides and try again?

https://github.com/OpenNebula/one/blob/master/share/install_gems/CentOS7/Gemfile.lock

n40lab:
Sure, I'll try and add it to documentation so it helps others. I'm closing this conversation so this PR is not shown anymore.

Thanks!

PR #214, opened 2017-03-12
Apply fix for "Requested VNC port already assgined to a VM" for undeployed VM

dfelsch:
As discussed in this forum thread [1], if the onedb fsck tool is run while there is a VM in state UNDEPLOYED, the fsck script mislabels the corresponding VNC port as being used which blocks resuming the VM later. The proposed fix to block only ports of VM in state RUNNING works for me (and others in the forum). Feel free to add further states that should block a VNC port (maybe SUSPENDED?).

[1] https://forum.opennebula.org/t/requested-vnc-port-already-assgined-to-a-vm/3194


dfelsch:
I would really appreciate if somebody would look (and hopefully merge) this simple one-line patch after more than 3 months...

jfontan:
Sorry for the delay. I'll check with @rsmontero which other states need to be taken into account.

jfontan:
I've reworked the change to take into account more states.

Thanks!

PR #223, opened 2017-03-16
Add support for Ceph DS with shared TM (Bug #4944)

lgrawet:
Hi,

This pull request intends to fix
[Bug #4944: tm/shared/clone: ceph disk clone is failing with shared system ds
](https://dev.opennebula.org/issues/4944) and more.

The pull request #160 from Anton Todorov (StorPool)  is required, otherwise ceph DS migration scripts are called resulting in total loss of VM files in shared system DS.

I also had to add "kvm-shared" to LIVE_DISK_SNAPSHOTS variable in
  /etc/one/vmm_exec/vmm_execrc
file to allow Ceph live disk snapshots while using shared system datastore.

Best regards,

Laurent

### DIFF
--- a/src/tm_mad/shared/clone
+++ b/src/tm_mad/shared/clone
@@ -67,10 +67,12 @@ unset i j XPATH_ELEMENTS
 while IFS= read -r -d '' element; do
     XPATH_ELEMENTS[i++]="$element"
 done < <(onevm show -x $VMID| $XPATH \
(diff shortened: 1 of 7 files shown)


rsmontero:
We are promoting full ceph mode and the ssh mode. This alleviates the need of setting up a shared FS together with Ceph. The full mode is the recommended setup. Future versions of OpenNebula will still 'silently' support the shared mode as is, but not additional features will be added to this shared mode.

THANKS for the commit in any case, it may be useful for people in the transition to the ceph mode.

PR #244, opened 2017-04-03
F #5068 set snapshot_id to the warning dialog on VM сnapshot revert

atodorov-storpool:
Probably there is more elegant solution, but with this patch the warning dialog POST proper JSON with snapshot_id parameter set.


atodorov-storpool:
resolved upstream

PR #245, opened 2017-04-05
vmm/kvm/{save,restore} hooks for save-pre and restore-post

atodorov-storpool:
qemu-kvm-ev works with checkpoint save/restore directly on block device ;)

I believe that should work with CEPH too...


atodorov-storpool:
And here are the scripts for our addon that are using this patch 
https://github.com/OpenNebula/addon-storpool/commit/2d5a3918529bbd231824ed6a0b5da10369f46628

PR #285, opened 2017-04-26
Update tm/failmigrate

atodorov-storpool:
When live migration fails, call other IMAGE_DS's failmigrate script(s) to clean up the destination host (detach disks, etc)


atodorov-storpool:
implmented

PR #289, opened 2017-04-27
Add OpenNebula systemd service dependency on MySQL

vholer:
References https://github.com/OpenNebula/docs/pull/93

On Debian-like systems, the MariaDB service name is still **mysql**. This small change adds soft dependency also on *mysql.service* into the systemd service script.


jfontan:
Looks good

PR #292, opened 2017-05-04
support rfc2307 groups in ldap

4nd3r:


4nd3r:
related: https://dev.opennebula.org/issues/4651

jfontan:
Thanks!

PR #299, opened 2017-05-08
Don’t create virsh domain snapshot metadata. REFS#5145

feldsam:
[Redmine bug](https://dev.opennebula.org/issues/5145)

Signed-off-by: Kristian Feldsam <feldsam@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.o
 *.pyc
 *.DS_Store
+.idea/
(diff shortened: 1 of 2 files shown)


jfontan:
I've cherry picket the commit 272a6037244f35b52d620cb391666dbcdf016b4a to master. It's not merged because the extra commit changing .gitignore.

Thanks!

PR #314, opened 2017-05-25
B #5171: Check if NIC is really absent after detach on KVM

vholer:


kvaps:
ref https://dev.opennebula.org/issues/5171

Hey wasn't this bug reported to libvirt team?

kvaps:
thanks!

PR #321, opened 2017-05-30
Fixed create support ticket severity custom field

feldsam:
Tested today towards zendesk and severity was not submitted correctly. I figure out, that we need to add severity value to tags also, because dropdown custom field use tags at the background.


rsmontero:
Thanks

PR #355, opened 2017-06-22
Sched devel

valzacc:
Added possibility to use custom priority in VM scheduling, e.g. given by the FairShare Scheduling service.


rsmontero:
Hi

Look great :+1:  I've a question. Right now the objects stored in a Pool are a map indexed by an int 
`std::map<int, ObjectXML*> `, I like the idea of reordering the map using the PRIO field, and inserting the VMs using PRIO. However when 2 VMS have the same priority the insert will[ fail ](https://github.com/OpenNebula/one/pull/355/files#diff-254fce62d33769465ddb944329d13c5aR118). Probably we could either use a multimap for PoolXML::objects or just sort the map using an ancillary container like a vector... (I like better the second approach). What do you think? I can take a closer look and come up with a proposal using the [SchedulerPolicy class](https://github.com/OpenNebula/one/blob/master/src/scheduler/include/SchedulerPolicy.h). Just returning the [VM/PRIO field will do, in here](https://github.com/OpenNebula/one/blob/master/src/scheduler/include/SchedulerPolicy.h#L82) .

It needs a bit of logic to work on VMs instead of Hosts, but that's the idea.

Make sense?

valzacc:
Hi Ruben,

thanks for your comment. 

Up to now we do have a setting which prevents having the same priority value (L133 in https://github.com/indigo-dc/one-fass/blob/master/src/pm/BasicPlugin.cc). 

Anyway, it is great if you have suggestions to integrate this functionality in ONE in a more effective way! I will discuss this with Sara and come back to this thread.

svallero:
Hi Ruben,
indeed we assumed that "whoever" provides the priority values should take care of having them unique. Our modification was the most straightforward we could think about without intruding too much in the ONE code, but clearly it's not very robust. Yes, your ideas make a lot of sense and personally I also prefer the second option. Look forward to have your improved version!

rsmontero:
Sara + Valentina

THANKS, I'll work on that. I'll update the issue so we can test it with the fair share scheduler

Cheers

rsmontero:
I've just pushed the changes into master.  https://github.com/OpenNebula/one/commit/1f571426ab257ea5993ec8828738e7b60f1049ef

Changes compared to the PR:

1. The attribute to store the computed VM priority is ``USER_PRIORITY`` this variable should be updated through the API (e.g. as in onevm update) (its xpath is ```/VM/USER_TEMPLATE/USER_PRIORITY```).   **Note**. I'm afraid this requires changes on fass scheduler (How do you update /VM/PRIO, is this a direct DB access?).

2. The use of  an USER_TEMPLATE  (pretty much as SCHED_RANK...) allows admins to set specific priorities to VM templates without using an external component.

3. Also, this attribute is set as a restricted, so only users (or processes) with oneadmin rights can modify this field. So fass needs access to a oneadmin capable credential (like the regular one scheduler), I guess this is not a problem.

4. Although changes could be done more straightforward, I kept the policy class abstraction so this  policy could be potentially combine with others. Priorities are scaled (i.e. priority 1 is always the highest) so the one shown in sched.log is scaled, in this way different policies can be used with a linear combination each computed priority.  

5. USER_PRIORITY supports floats, just in case it is useful in fass.

This is now in master, and will be released in 5.4 (internal testing is undergoing). If you have the opportunity to test it with fass, before the final release (just a week, sorry for that) we could include any bug fix or minor improvement.

Cheers

valzacc:
Hi Ruben,
thanks for your work! 

You can go ahead with your tests before 5.4. There are some small things which we need to change in fass and we'll do our best to try out the master before your release. In case you don't hear from us, we did not make it in time. Anyway, there are no problems from our side with your PR! Some comments:

1. here we have to change the way we write the xml, but it is a minor thing. We do not access the one db, since we do communicate only with your scheduler, via xmlrpc calls.
2. ok
3. no problem, fass already knows the oneadmin credentials
4. just one question: is the rescaling done directly by one? 
5. very good, it is indeed useful!

Thanks again and cheers,
Valentina and Sara

rsmontero:
GREAT THANKS!. It is now in master for the 5.4RC.

About 5. Yes the scaling is done by OpenNebula scheduler, so you don't need to do anything in fass. Just note that now in sched.log we print each VM priority in the dispatch list. That priority is scaled and won't be the same as the value of USER_PRIORITY in the Template.

valzacc:
Hi Ruben,

We have tested FASS with ONE 5.4 it and there are no problems. 
New version in master: https://github.com/indigo-dc/one-fass works with ONE 5.4, while version v1.1 works with previous versions of ONE.
We close now the pull request and thanks a lot,

Valentina and Sara

rsmontero:
Excellent :)  THANKS!

PR #362, opened 2017-06-29
Add support for CEPH_KEY datastore attribute to be able to authenticate against a ceph cluster

hydro-b:
See https://dev.opennebula.org/issues/5208


rsmontero:
Thanks. I've modified the export command to generate the URL cleaner. Also the other side (import) needed to be modified to get the key. I didn't have access to Ceph cluster so I couldn't try the export to marketplace, though

rsmontero:
BTW, link to the commit
https://github.com/OpenNebula/one/commit/dde9cf317d70ee51f95796c536604f146d62bf8a

PR #367, opened 2017-06-30
xpath.rb fix crash when querying empty-element tag

atodorov-storpool:
by returning empty string.
Handling both `<test/>` and `<test></test>`
test pattern:
```
echo -e "<A>\n<B>b</B>\n<C/>\n<D></D><E>e</E></A>\n"| /var/lib/one/remotes/datastore/xpath.rb /A/B /A/C /A/D /A/E | while IFS= read -r -d '' e; do echo "'$e'"; done
```


jfontan:
Thanks!

PR #403, opened 2017-07-18
changed around hybrid conf files

Semedi:


jmelis:
The asterisk is invalid. Can you resend the patch without it? something like

.gsub(/[^a-zA-Z0-9_.]/,"")

PR #417, opened 2017-07-26
R #5283 Added scroll bar in Sunstone VM Log

abelCoronado93:


dann1:
Thanks a lot for the quick reaction :+1:

dann1:
Hello again, can you enable the default view to be at the end of the log file. When you press the refresh button or another button the scroll bar goes to the beginning. Instead it would be useful for it to be at the end because it contains the latest entries.

abelCoronado93:
I've opened an issue: https://github.com/OpenNebula/one/issues/1617

Thanks for the report!

PR #424, opened 2017-08-04
Fix vxlan device creation

bewie:
Since the upgrade to opennebula 5.4, the creation of vxlan doesn't work.
In the commit fe84d376acb079c1c1a357e9e69834fcbba2502d the module `VNMMAD::VNMNetwork::IPv4` has been removed and `IPAddr` class used.
But on merge 1b28fbe354f84177ac46065d7fb22c880e288127 `VNMMAD::VNMNetwork::IPv4` is used again but without the module declaration in `lib/address.rb`


rsmontero:
THANKS!!!!

PR #425, opened 2017-08-11
Send unsolicited ARP packet to update IP across switches

sulaweyo:
The arping parameter needs to be -U instead of -A as an unsolicited ARP packet is needed to update the ARP tables in switches. Without that the switch of the IP to a different host is not recognized and requests to the floating IP do not reach the correct node.

Bug report https://dev.opennebula.org/issues/5309

### DIFF
--- a/share/hooks/raft/vip.sh
+++ b/share/hooks/raft/vip.sh
@@ -17,7 +17,8 @@ fi
 case $ACTION in
 leader)
     sudo ip address add $IP dev $INTERFACE
-    sudo arping -c 5 -U -I $INTERFACE ${IP%%/*}
+    sudo arping -c 3 -U -I $INTERFACE ${IP%%/*}
+    sudo arping -c 3 -UA -I $INTERFACE ${IP%%/*}
     ;;
 
 follower)
--- a/share/hooks/raft/vip.sh
+++ b/share/hooks/raft/vip.sh
@@ -17,7 +17,7 @@ fi
 case $ACTION in
 leader)
     sudo ip address add $IP dev $INTERFACE
-    arping -c 5 -A -I $INTERFACE ${IP%%/*}
+    arping -c 5 -U -I $INTERFACE ${IP%%/*}
     ;;
 
 follower)
--- a/share/hooks/raft/vip.sh
+++ b/share/hooks/raft/vip.sh
@@ -17,7 +17,7 @@ fi
 case $ACTION in
 leader)
     sudo ip address add $IP dev $INTERFACE
-    arping -c 5 -U -I $INTERFACE ${IP%%/*}
+    sudo arping -c 5 -U -I $INTERFACE ${IP%%/*}
     ;;
 
 follower)


share/hooks/raft/vip.sh
```diff
@@ -17,7 +17,7 @@ fi
 case $ACTION in
 leader)
     sudo ip address add $IP dev $INTERFACE
-    arping -c 5 -U -I $INTERFACE ${IP%%/*}
+    sudo arping -c 5 -U -I $INTERFACE ${IP%%/*}
```
vholer: Sudo shouldn't be needed as the arping binary (should) have the file capabilities set:

```
$ getcap $(which arping)
/usr/sbin/arping = cap_net_raw+p
```

But, it can be better to have sudo there.

Just curious -- what platform, filesystem,  (SELinux/AppArmor?) do you use and can you copy&paste the executed command and error output? Not necessary, thank you.

hydro-b:
Hi there. I saw this pull request passing by. I agree with @sulaweyo that the "-U" parameter should be used. You can however also add the "-A" option. You will get different behaviour. First one does a "who-has with an answer", the second version (with -UA) gives a truely gratoutouis arp. See examples here:

-U option only: 

1502882189.531272 Out 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Request who-has 172.17.158.10 (ff:ff:ff:ff:ff:ff) tell 172.17.158.10, length 28
1502882190.531832 Out 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Request who-has 172.17.158.10 (ff:ff:ff:ff:ff:ff) tell 172.17.158.10, length 28
1502882191.532047 Out 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Request who-has 172.17.158.10 (ff:ff:ff:ff:ff:ff) tell 172.17.158.10, length 28
1502882192.532255 Out 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Request who-has 172.17.158.10 (ff:ff:ff:ff:ff:ff) tell 172.17.158.10, length 28
1502882193.532646 Out 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Request who-has 172.17.158.10 (ff:ff:ff:ff:ff:ff) tell 172.17.158.10, length 28

-UA option:

1502882607.884242   B 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Reply 172.17.158.10 is-at 52:54:00:d5:76:bf, length 28
1502882608.884208   B 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Reply 172.17.158.10 is-at 52:54:00:d5:76:bf, length 28
1502882609.884600   B 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Reply 172.17.158.10 is-at 52:54:00:d5:76:bf, length 28
1502882610.884556   B 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Reply 172.17.158.10 is-at 52:54:00:d5:76:bf, length 28
1502882611.884745   B 52:54:00:d5:76:bf ethertype ARP (0x0806), length 44: Reply 172.17.158.10 is-at 52:54:00:d5:76:bf, length 28

Maybe we can have both options active after another. 3 arping's with "-U" only, 3 arpings with "-UA". Some equipment might update arptable based on first 3 arpings, other only on the second series of arpings (if at all).

vholer:
Hi @hydro-b , I was considering a change with both -U and -A:
https://github.com/OpenNebula/one/pull/426/files
as this is that Linux HA project suggests 

https://wiki.wireshark.org/Gratuitous_ARP:
Note that some devices will respond to the gratuitous request and some will respond to the gratuitous reply. If one is trying to write software for moving IP addresses around that works with all routers, switches and IP stacks, it is best to send both the request and the reply.
http://hg.linux-ha.org/heartbeat-STABLE_3_0/file/73e18422f657/tools/send_arp.libnet.c#l239

hydro-b:
-    arping -c 5 -A -I $INTERFACE ${IP%%/*}
+    for i in $(seq 5); do
+        arping -c 1 -U -I $INTERFACE ${IP%%/*}
+        sleep 1
+        arping -c 1 -A -I $INTERFACE ${IP%%/*}
+        sleep 1
+    done

^^ I would change that to:

-    arping -c 5 -A -I $INTERFACE ${IP%%/*}
+    for i in $(seq 5); do
+        arping -c 2 -U -I $INTERFACE ${IP%%/*}
+        sleep 1
+        arping -c 2 -UA -I $INTERFACE ${IP%%/*}
+        sleep 1
+    done

You need both -UA together, and  can use -U alone. I would send two arp's in case one gets lost (I have seen that happening).

hydro-b:
pff, markdown ... completely screwed that up. Sorry
```
-    arping -c 5 -A -I $INTERFACE ${IP%%/*}
+    for i in $(seq 5); do
+        arping -c 2 -U -I $INTERFACE ${IP%%/*}
+        sleep 1
+        arping -c 2 -AU -I $INTERFACE ${IP%%/*}
+        sleep 1
+    done
```

sulaweyo:
I just used -U as this was actually enough to get it working and nobody sees the response anyway but i have nothing against -UA and even the loop would be ok. As the leader can switch pretty fast via raft the count of 5 is pretty high imho and -c 1 would actually be enough.

Whats important though is the sudo in front of it (or setuid on arping)

share/hooks/raft/vip.sh
sulaweyo: I'm running on arch and there is no setuid flag on arping so i added the sudo as it's already in for the ip command

vholer:
On Fri, Aug 18, 2017 at 1:33 PM Sledge Sulaweyo <notifications@github.com>
wrote:

> I just used -U as this was actually enough to get it working and nobody
> sees the response anyway but i have nothing against -UA and even the loop
> would be ok. As the leader can switch pretty fast via raft the count of 5
> is pretty high imho and -c 1 would actually be enough.
>
> Whats important though is the sudo in front of it (or setuid on arping)
>
I'm looking on the "-UA" option, and it's ambiguous. It's not clear at the
first sight what is it doing. After looking into the code:

https://github.com/iputils/iputils/blob/master/arping.c#L1030
... both -U and -A set the unsolicited mode (unsolicited++)

https://github.com/iputils/iputils/blob/master/arping.c#L281
... but in the execution, only -U **or** -A is taken into the account

So it still looks to me better using -U or -A for better readibility.
Aren't you using some other implementation of the arping?

I have updated my branch so it also uses and configures sudo. I believe, it
should be fine now. https://github.com/OpenNebula/one/pull/426/files

hydro-b:
I'll test your patch in our test environment and let you know what arping's I see. Using "arping" from iputils.

vholer:
Thank you @hydro-b and @sulaweyo for your code and precious comments, we are going to merge PR #426, as there are all necessary changes incl. sudoers.

PR #427, opened 2017-08-23
Ldap filter can use `;` character

dudoslav:
This change will allow user_field to contain characters such as `;`. This is useful because some of ldap attributes look like `login;organization`.


jfontan:
Thanks! merged to master and one-5.4 branches

PR #510, opened 2017-09-28
B #5416: Import vCenter template discard linked clones value

abelCoronado93:


jfontan:
Added also to one-5.4

PR #528, opened 2017-10-13
LUKS support for images (Feature 5402)

laurencegill:
In regards to https://dev.opennebula.org/issues/5402

I have tested this works with images from a template and also when attaching images to a running VM.

Happy to add the documentation for working with persistent luks disks.  I have tested creating new disks and converting existing persistent images (requires qemu 2.10).

I think to have ephemeral luks images some kind of secret management would be required in opennebula to generate and add the libvirt secret to the hypervisor with the correct volume name?


laurencegill:
Is this likely to get merged as I  have to manually apply the patch each time we upgrade?  I can write the full documentation for users...

Cheers

laurencegill:
https://github.com/OpenNebula/one/issues/1345#issuecomment-564035506

laurencegill:
rebased onto new branch, new PR submitted with correct commit syntax message

PR #545, opened 2017-10-25
extend attributes sent to network drivers

albertoflorez:


albertoflorez:
https://dev.opennebula.org/issues/5493

jmelis:
Thanks a lot!!

PR #573, opened 2017-11-14
sudo for dd doesn't seem necessary

Brennwert:
I can't find any call of dd with sudo in the whole project. While having the opportunity to escalate permissions with dd it's maybe a good idea to remove it from sudoers.
Am I missing something?

### DIFF
--- a/share/sudoers/sudo_commands.rb
+++ b/share/sudoers/sudo_commands.rb
@@ -20,7 +20,7 @@ require "erb"
 
 
 CMDS = {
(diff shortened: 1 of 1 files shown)


vholer:
Thank you, `dd` was already removed by commit https://github.com/OpenNebula/one/commit/639c866798a74c7a3e5f1c51b35ee52f2c56db91, but I forgot on this script. I have prepared another PR with your commit and commit adding some missing tools. Thank you!

PR #577, opened 2017-11-16
sudo for mkfs also questionable and mkswap partly missing

Brennwert:
mkswap was missing in openSUSE sudoers-distributionfile and in sudo_commands.rb

Also similar to dd I was reviewing the necessity of sudo for mkfs -  wasn't able to find a line where it's called with elevated privileges. But maybe I'm misunderstanding concepts here.


vholer:
Since there is a conflict, I have created a separate issue https://github.com/OpenNebula/one/issues/3046 to make the proposed changes + add few more cleanups related to the LXD sudoers management. Closing this PR for now, thank you!

PR #580, opened 2017-11-16
Fix user login via LDAP (refs #1489)

fpeyraud:
User parameter was not a table anymore, and first parameter of 'is_in_group' should be the user's dn

### DIFF
--- a/src/authm_mad/remotes/ldap/authenticate
+++ b/src/authm_mad/remotes/ldap/authenticate
@@ -76,7 +76,7 @@ begin
     end
 
     if server_conf[:group]
-        if !ldap.is_in_group?(user_group_name, server_conf[:group])
+        if !ldap.is_in_group?(user, server_conf[:group])
             STDERR.puts "User #{user} is not in group #{server_conf[:group]}"
             next
         end
--- a/src/authm_mad/remotes/ldap/ldap_auth.rb
+++ b/src/authm_mad/remotes/ldap/ldap_auth.rb
@@ -155,7 +155,7 @@ class OpenNebula::LdapAuth
     end
 
     def is_in_group?(user, group)
-        username = user.first.force_encoding(Encoding::UTF_8)
+        username = user.force_encoding(Encoding::UTF_8)
         result=@ldap.search(
                     :base   => group,
                     :attributes => [@options[:group_field]],


vholer:
Hi @fpeyraud, this proposed change is not valid (although it may fix problem in your env.). I'll explain in the issue #1489. Correct fix is in PR #1523.

vholer:
Explanation better here. Having following LDAP structure with a (POSIX) user account and group:

```yaml
dn: cn=cloud,ou=Group,dc=example,dc=org
objectClass: top
objectClass: posixGroup
cn: cloud
gidNumber: 30000
memberUid: ldapuser1

dn: uid=ldapuser1,ou=People,dc=example,dc=org
objectClass: top
objectClass: account
objectClass: posixAccount
objectClass: shadowAccount
cn: ldapuser1
uid: ldapuser1
uidNumber: 30000
gidNumber: 30000
homeDirectory: /home/ldapuser1
loginShell: /sbin/nologin
gecos: OpenNebula LDAP user
shadowLastChange: 17058
shadowMin: 0
shadowMax: 99999
shadowWarning: 7
```

For group membership checks, you need to specify attributes on both sides which define the user-group relationship. In this case, the relevant OpenNebula LDAP configuration (http://docs.opennebula.org/5.4/deployment/authentication_setup/ldap.html#configuration) parameters are:

* **:group_field**: 'memberUid'     *# ... attribute in the group entry which references to the user*
*  **:user_group_field**: 'uid'         *# ... attribute in the user entry which identifies the user in the group entry*

```diff
-        if !ldap.is_in_group?(user_group_name, server_conf[:group])
+        if !ldap.is_in_group?(user, server_conf[:group])
```

The proposed patch fallbacks to using just user login for user-group membership matching, which is a valid for the example tree above, but not in general. We really need to use `use_group_name` (value of the `user_group_field`), we can't assume everybody is just using `login` there.

For example, the following example of the ActiveDirectory tree:

```yaml
dn: CN=cloud,CN=Groups,DC=example,DC=org
...
member: CN=test test,CN=Users,DC=example,DC=org
member: CN=user A,CN=Users,DC=example,DC=org
member: CN=user B,CN=Users,DC=example,DC=org
member: CN=user C,CN=Users,DC=example,DC=org

dn: CN=test test,CN=Users,DC=example,DC=org
...
cn: test test
sn: test
distinguishedName: CN=test test,CN=Users,DC=example,DC=org
```

... have users' DN for group membership. The PR would break all these environments as it would check group `member` attributes always just for the `login`.

Issue #1489, opened 2017-11-20
Cannot authenticate via LDAP with :group set


OpenNebulaProject:
---

Author Name: **Florent Peyraud** (Florent Peyraud)
Original Redmine Issue: 5559, https://dev.opennebula.org/issues/5559
Original Date: 2017-11-16

---

Hi
When trying to authenticate with LDAP using the :group parameter set to restrict the users allowed to log in, it fails with following logs

```
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 Command execution fail: /var/lib/one/remotes/auth/ldap/authenticate florent.peyraud - ****

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: Command execution fail: /var/lib/one/remotes/auth/ldap/authenticate florent.peyraud - ****
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 Trying server server 1

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: Trying server server 1
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 Exception raised authenticating to LDAP

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: Exception raised authenticating to LDAP
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 #<NoMethodError: undefined method `force_encoding' for nil:NilClass>

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: #<NoMethodError: undefined method `force_encoding' for nil:NilClass>
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 /usr/lib/one/ruby/opennebula/ldap_auth.rb:158:in `is_in_group?'

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: /usr/lib/one/ruby/opennebula/ldap_auth.rb:158:in `is_in_group?'
Thu Nov 16 17:34:00 2017 [Z0][AuM][D]: Message received: LOG I 8 /var/lib/one/remotes/auth/ldap/authenticate:79:in `block in <main>'

Thu Nov 16 17:34:00 2017 [Z0][AuM][I]: /var/lib/one/remotes/auth/ldap/authenticate:79:in `block in <main>'
[...]

PR #1523, opened 2017-11-25
B #1489: Fix LDAP group membership checks with custom 'user_group_field'

vholer:


vholer:
A patch fixes the LDAP group membership checking for custom `user_group_field` (it worked only for `dn`, `memberOf` and a configured `:user_field`). Without the patch, relevant `user_group_field` attribute isn't fetched for any user, the variable `user_group_name` is left undefined and naturally leads to the authentication failure.

PR #1568, opened 2017-12-06
Extension of ceph drivers to support ceph erasure coding

Mamut3D:
I have extended ceph drivers to support cephp pools with erasure coding (released in Luminous http://ceph.com/community/new-luminous-erasure-coding-rbd-cephfs/)

Erasure coding feature is enabled by adding attribute EC_POOL  the ceph datastore in nebula.
EC_POOL have to contain the name of the erasure coded ceph pool.
EC_POOL can be added to system and/or image datastore to enable erasure coded RBDs only in the selected datastore. 

Extension is backwards compatible


src/tm_mad/ceph/clone
```diff
@@ -54,6 +54,7 @@ SRC_PATH=`arg_path $SRC`
 
 DISK_ID=$(echo $DST|awk -F. '{print $NF}')
 RBD_DST="${SRC_PATH}-${VM_ID}-${DISK_ID}"
+DST_DS_ID=$(echo $DST | sed s#//*#/#g | awk -F/ '{print $(NF-2)}')
```
jmelis: Maybe better like this:
`echo $DST | $SED 's#//*#/#g' | $AWK -F/ '{print $(NF-2)}'`

src/tm_mad/ceph/clone
```diff
@@ -78,6 +79,19 @@ CEPH_CONF="${XPATH_ELEMENTS[j++]}"
...
+
+unset i j XPATH_ELEMENTS
+
+while IFS= read -r -d '' element; do
+        XPATH_ELEMENTS[i++]="$element"
+      done < <(onedatastore show -x $DST_DS_ID | $XPATH \
+                            /DATASTORE/TEMPLATE/EC_POOL)
+
+EC_POOL="${XPATH_ELEMENTS[j++]}"
+
```
jmelis: No need if `INHERIT_DATASTORE_ATTR  = "EC_POOL"` is added to `oned.conf`

src/tm_mad/ceph/mkimage
```diff
@@ -83,6 +83,23 @@ CEPH_CONF="${XPATH_ELEMENTS[j++]}"
...
+        XPATH_ELEMENTS[i++]="$element"
+      done < <(onedatastore show -x $DSID | $XPATH \
+                            /DATASTORE/TEMPLATE/EC_POOL)
+
+EC_POOL="${XPATH_ELEMENTS[j++]}"
+
+#-------------------------------------------------------------------------------
+# Make the new image (file-based)
+#-------------------------------------------------------------------------------
+
```
jmelis: same as above

src/tm_mad/ceph/clone
Mamut3D: I have written it consistently with the rest of the script - `awk` is used on line 55 `DISK_ID=$(echo $DST|awk -F. '{print $NF}')`

Should I change it there as well? What about other ceph driver scripts?

src/tm_mad/ceph/clone
Mamut3D: I have some issues with the change `INHERIT_DATASTORE_ATTR`:

The good:
It will be faster, since it won't need to call additional `onedatastore show`
Configuration options will be simpler

The bad:
In my original implementation EC_POOL setting is taken into account separately for  `image datastore` and `system datastore`, so it allows the following combinations
- you can have `regular replicated` images in `image datastore`, which can be deployed as `erasure coding` image to `system datastore`
- you can have `erasure coding` images in `image datastore`, which can be deployed as  `regular replicated` images in `system datastore`

The ugly:
Your proposed solution with `INHERIT_DATASTORE_ATTR` will always respect only `image datastore` EC_POOL and use it on `system datastore`

Which options suits you better?

src/tm_mad/ceph/clone
jmelis: We should definitely change that, but since we need to do it, you can leave it as is.

src/tm_mad/ceph/clone
jmelis: I stand corrected. Your proposal makes more sense. Thanks for taking the time to explain it to me.

jmelis:
So, from my part I would like to merge this PR, but before doing so I would like to create tests in our internal infrastructure and try everything. This may take sometime but I'm assigning it a high priority.

Would you consider doing a PR for the docs in master?

Thanks again!

vholer:
> Add more commits by pushing to the ceph_ec branch on Mamut3D/one.

Wow, I didn't know I can push into someone's fork! This was accidental.

Issue #1531, opened 2017-11-27
opennebula should not import uplink portgroups


Semedi:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [X] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Sunstone
- [ ] Storage
- [ ] Scheduler
- [X] vCenter

## Description
Try to import any network from opennebula, you will be able to import every distributed port group and their uplink portgroup associated.

### Expected Behavior
opennebula should not import uplink portgroups

PR #1638, opened 2018-01-08
B #1531: vcenter_driver should ignore uplink port groups

Semedi:


Semedi:
no, this change applies only master due to it needs the vcenter importing networks rework.

If we want to have this change applied to 5.4 we need to develop new code for that branch.

Issue #1371, opened 2017-11-20
Migrate VMs between different vCenter clusters


OpenNebulaProject:
---

Author Name: **OpenNebula Systems Support Team** (OpenNebula Systems Support Team)
Original Redmine Issue: 5438, https://dev.opennebula.org/issues/5438
Original Date: 2017-10-05

---

OpenNebula should provide this functionality, which is already present in vCenter.

tinova:
[ ] Testing
 [ ] Documentation: Explain requirements cluster migration: shared networks and datastores between OpenNebula clusters

PR #1643, opened 2018-01-10
F #1371: vcenter vm cluster migration

Semedi:


Semedi:
we need the mv script to take benefit of the oneUtil bash library.

So to be able to do the same thing with mv we need to develop the same library on ruby.

PR #1652, opened 2018-01-16
Fix performance issue in sunstone / mysql

nierob:
Sunstone often queries for running VMs. Over time the query became
quite expensive because it scans for all VMs that were ever created.
For our case in which we had more then 600k done machines, the query
time was longer then 25s, causing sunstone to start pondering about
it's db connection and freezing.

This patch tries to mitigate the problem by installing an index on
vm_pool.state column, so running VMs could be found in almost no
time.

### DIFF
--- a/src/onedb/database_schema.rb
+++ b/src/onedb/database_schema.rb
@@ -57,7 +57,7 @@ class OneDBBacKEnd
             vm_pool: "oid INTEGER PRIMARY KEY, name VARCHAR(128), " <<
                 "body MEDIUMTEXT, uid INTEGER, gid INTEGER, " <<
                 "last_poll INTEGER, state INTEGER, lcm_state INTEGER, " <<
-                "owner_u INTEGER, group_u INTEGER, other_u INTEGER",
+                "owner_u INTEGER, group_u INTEGER, other_u INTEGER, index(state)",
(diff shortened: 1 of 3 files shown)


nierob:
Please review carefully I haven't executed any test and in addition I'm not a ruby expert.

nierob:
Ping?

rsmontero:
Thanks, @nierob. Let me be sure I understand this.

So we currently perform queries based on the state column. In your setup, this queries based on STATE (e.g. STATE <> 6 ) are slow because a high number of VMs. This patch is to install an index for this column and so to speed up the process, right?

nierob:
Yes, that is correct. Whenever one refresh VM view, a sql query is done on vm_pool table. The view shows only running integrations therefore the sql query contains "where state<>6", which without an index, involves a full table scan, which is very slow.

rsmontero:
Redo some of this PR in this commit, basically to homogenize with the other indexes of logdb table.

https://github.com/OpenNebula/one/commit/37e14b23f5947bb2e950ab21e86fbf34ad2263f5

rsmontero:
BTW, in recent verions of opennebula you can use purge the DB with purge-done and purge-history.

nierob:
I tried purge-done but it exploded with a ruby exception, sadly I haven't captured it and I have chicken out to run it again :-) I will report it if I see it again. Thanks.

PR #1714, opened 2018-02-05
B #1713: DS quotas shouldn't show the 'system' type

abelCoronado93:


src/sunstone/public/app/utils/quotas/quota-widgets.js
```diff
@@ -506,6 +506,8 @@ define(function(require) {
           emptyValue: true
         });
 
+      $(".ds_select .resource_list_select option[elem_id='0']", context).remove();
```
tinova: does this cover all system DSs, or just DS with ID = 0?

abelCoronado93: Just '0: system' DS

Issue #1538, opened 2017-11-28
Import multi-cluster networks


tinova:
# Enhancement Request
## Description

Networks and distributed switches in vCenter can span across multiple vCenter clusters, which are modeled as OpenNebula clusters.

At network import time, if more than one vCenter cluster is imported in OpenNebula, and the network belongs to these clusters, an option to import the virtual network into the different OpenNebula clusters should be presented

## Use case

Full use of Networks that span across different vCenter clusters

## Interface Changes

* onevcenter import 
* Sunstone network from vcenter import dialogs

Semedi:
Whe should improve this feature following the next use case:

## Use case
* You need to import a net that belongs to two or more clusters.
* You dont have all of these cluster imported into ONE.
* ONE should ask you about import the other clusters

tinova:
# Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

tinova:
[ ] Sunstone network import in multiple cluster will be treated in #1653
 [ ] Hooks for network creation won't support multi cluster creation
 [ ] Improve hooks for network creation opportunistically
 [ ] Merge multi network creation functionality (Sergio's knows)

tinova:
Sunstone missing

rsmontero:
Sunstone network import in multiple cluster will be treated in #1653

PR #1718, opened 2018-02-07
F #1538: import virtual_machines and templates networking update

Semedi:
F #1538: create_one_network created, it handles cluster selection

### DIFF
--- a/src/vmm_mad/remotes/lib/vcenter_driver/importer.rb
+++ b/src/vmm_mad/remotes/lib/vcenter_driver/importer.rb
@@ -641,9 +641,9 @@ def self.import_networks(con_ops, options)
                 if !use_defaults
                     message = false
                     print_str =  "\n  * Network found:\n"\
-                                 "      - Name                     : \e[92m#{n[:name]}\e[39m\n"\
-                                 "      - Type                     : #{n[:type]}\n"\
-                                 "      - Vcenter Clusters(host id): "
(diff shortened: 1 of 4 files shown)


src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -614,6 +601,7 @@ def get_vcenter_nics
                 else
                     network = device.backing.network
                 end
+                nic[:refs] = network.host.map {|h| h.parent._ref}
```
tinova: This should have error checking in case h.parent or h.parent._ref does not exist

src/vmm_mad/remotes/lib/vcenter_driver/importer.rb
```diff
@@ -801,32 +801,12 @@ def self.import_networks(con_ops, options)
...
                 n[:one] << ar_str
+                n[:clusters][:one_object] = n[:one]
 
-                one_vn = VCenterDriver::VIHelper.new_one_item(OpenNebula::VirtualNetwork)
-                if n[:clusters] #Distributed Port Group
-                    done = []
-                    for i in 0..n[:clusters][:refs].size-1
-                        cl_id = n[:clusters][:one_ids][i]
-                        next if cl_id == -1 || done.include?(cl_id)
+                one_vn = VCenterDriver::Network.create_one_network(n[:clusters])
```
tinova: object passed to create_one_network should be explicitly created to improve readibility

PR #1850, opened 2018-03-12
Add support for PowerPC CPUs

sulaweyo:
This minor adjustment allows the detection of PPC CPUs by the standard KVM driver.
I have 2 POWER8 machines that i use as KVM host with this change in place.

The only thing needed besides that is a correct node selection and the definition of EMULATOR in the VM template

### DIFF
--- a/src/im_mad/remotes/kvm-probes.d/cpu.sh
+++ b/src/im_mad/remotes/kvm-probes.d/cpu.sh
@@ -16,14 +16,20 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
+MODEL=''
+
 if [ -f /proc/cpuinfo ]; then
-    MODELNAME_EXP="model name"
-    if (uname -m | grep -Eq ^ppc64); then
-        MODELNAME_EXP="model"
-    fi
+    for NAME in 'model name' 'cpu'; do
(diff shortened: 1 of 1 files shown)


vholer:
Hello @sulaweyo , can you please provide us your content of `/proc/cpuinfo` ? Just to check. Thank you!

vholer:
Hallo @sulaweyo, thank you. It looks to me the CPU model is under the key `cpu`. The `model name` is the machine model. In your case, I would expect the probe to return:

```
MODELNAME="POWER8E (raw), altivec supported"
```

Can you please check this  https://gist.github.com/vholer/4cdfb3c4535d6f6ed29a4d78c0adca40, and eventually update your PR? Thank you!

sulaweyo:
Both hosts show up like that in ON
MODELNAME | 8247-22L
Which is expected as the model field is returned and not the cpu field of each core

vholer:
Yes, that's the state after your change, but you are showing the machine model, not CPU model. So, if you take the updated `cpu.sh` from https://gist.github.com/vholer/4cdfb3c4535d6f6ed29a4d78c0adca40, it should return the CPU model even for PPC.

sulaweyo:
Sorry for misreading it.
The gist cpu.sh indeed returns MODELNAME="POWER8E (raw), altivec supported". I think that provides less usable information than the model but if you prefer that output i'll adjust the pull request accordingly.

Issue #1872, opened 2018-03-15
Let ACL VDC rules to be configurable by admins.


OpenNebulaProject:
# Enhancement Request
## Description

Let users for example to list their own VDCs requires to grant VDC use to them. Currently this permission is not granted. Also for some setups it may be interesting to let the users check the assigned clusters (and their characteristics like number of hosts) to the VDC.

Also when a cluster is added to a VDC a set of automatic rules are generated to let users access all cluster resources. Sometimes it's better to fine grain tune these rules.

## Use case

Fine grain control on ACL rules for VDCs

## Interface Changes

 
oned.conf should include a list of rules to add:

VDC_DEFAULT_ACLS = "defalt acls generated to access VDC resources"
VDC_CLUSTER_DEFAULT_ACLS = "default acl generated when a cluster is added to a VDC"

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

kvaps:
@rsmontero would it work in case of specify in VDC template?

kvaps:
Probably don't

PR #1904, opened 2018-03-27
F #1872: Added new attributes to define default acls

juanmont:


include/Vdc.h
```diff
@@ -107,6 +107,8 @@ class ResourceSet
      */
     void del_rule(int group_id, int zone_id, int id);
 
+    void insert_defult_rules(string default_vdc_acl, PoolObjectSQL::ObjectType type);
```
rsmontero: This can be moved to the private section.

It needs comments

src/vdc/Vdc.cc
```diff
@@ -397,34 +397,61 @@ int Vdc::del_group(int group_id, string& error_msg)
...
+{
+    vector<string> vdc_acl;
+    vector<string>::const_iterator it;
+    long long op = AuthRequest::NONE;
+
+    vdc_acl = one_util::split(default_vdc_acl, '+', true);
+
+    for (it = vdc_acl.begin(); it != vdc_acl.end(); ++it)
+    {
+        if ( *it != "" && *it != "-" && AuthRequest::str_to_operation(*it) != AuthRequest::NONE )
```
rsmontero: AuthRequest::str_to_operation(*it)  is called twice maybe it'd be better to saved in a variable:

AuthRequest::Operation user_op =  AuthRequest::str_to_operation(*it);

and then use that

src/vdc/Vdc.cc
```diff
@@ -397,34 +397,61 @@ int Vdc::del_group(int group_id, string& error_msg)
...
-        break;
-
-        // @<gid> DATASTORE/#<dsid> USE #<zone>
         case PoolObjectSQL::DATASTORE:
-            rules.insert(make_pair(PoolObjectSQL::DATASTORE, AuthRequest::USE));
-            xml_name = "DATASTORE";
-        break;
+            xml_name = PoolObjectSQL::type_to_str(type);
+
+            Nebula::instance().get_configuration_attribute("DEFAULT_VDC_" + xml_name + "_ACL",default_vdc_acl);
```
rsmontero: 80 chars length?

PR #1912, opened 2018-03-28
Added go example and modify docker plugin location

christian7007:


Andromedae:
Hi guys, @christian7007 @tinova,

Is this (ie. one/oca/go) where the one go bindings are going to be maintained from now on ?
For third party users should we use this repo as opposed to OpenNebula/goca ?

Thank you !

tinova:
correct, goca has been integrated into OpenNebula, so the bindings will be maintained in one/oca/go

Issue #1797, opened 2018-02-27
System DS migration within the same host removes VM from the host inventory


vholer:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [ ] 5.4.4
- [ ] 5.4.5
- [ ] 5.4.6
- [ ] 5.4.7
- [X] 5.4.8
- [X] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [X] Storage & Images
- [ ] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## Description
If VM is migrated between system DS within the same host, the VM is migrated successfully, but the reference between HOST-VM is lost. Experienced with system DS of type SSH/LVM, but expect to be a general problem. If migrated between different hosts, the (new)HOST-VM reference is correctly updated, preserved.

### Expected Behavior
No matter where we migrate, the reference isn't lost.

## How to reproduce
0. have cleaned environment, no VMs running
1. have 2 system DS of same type
2. deploy a VM
3. migrate a VM within the same host to the other system DS
```
$ onevm show 17 | egrep ^HOST
HOST                : centos7-kvm-ssh-3ca29-2.test
```

4. check the onehosts for zero allocation

```
$ onehost list
  ID NAME            CLUSTER   RVM      ALLOCATED_CPU      ALLOCATED_MEM STAT  
   0 centos7-kvm-ssh default     0       0 / 100 (0%)   0K / 992.3M (0%) on    
   1 centos7-kvm-ssh default     0       0 / 100 (0%)   0K / 992.3M (0%) on   
```

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

vholer:
[...]

PR #1916, opened 2018-04-03
B #1797: Checked for the same host

juanmont:
cherry-pick to one-5.4

### DIFF
--- a/src/lcm/LifeCycleStates.cc
+++ b/src/lcm/LifeCycleStates.cc
@@ -73,7 +73,8 @@ void  LifeCycleManager::save_success_action(int vid)
 
         if ( vm->get_hid() != vm->get_previous_hid() )
         {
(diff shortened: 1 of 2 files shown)


src/lcm/LifeCycleStates.cc
```diff
@@ -71,7 +71,10 @@ void  LifeCycleManager::save_success_action(int vid)
 
         vm->get_requirements(cpu, mem, disk, pci);
 
-        hpool->del_capacity(vm->get_previous_hid(),vm->get_oid(),cpu,mem,disk,pci);
```
rsmontero: This del_capacity comes from add_capacity in LifeCycleActions::244. This add action is always performed  (no check current==previous), so capacity will be added twice.

Tests need to include this check also (host capacity is preserved)

PR #1917, opened 2018-04-03
Driver fs_lvm. method resize has been implemented

kazh000:
I have implemented the method "resize" for the fs_lvm storage driver. Could you accept one.


andir:
Is there any reason to not include this?

vholer:
OK from my side.

PR #1958, opened 2018-04-11
load function for addons implement

dcorral3:


src/sunstone/sunstone-server.rb
```diff
@@ -350,6 +350,60 @@ def destroy_session
         session.clear
         return [204, ""]
     end
+
+    def load_addons
+        puts "Loading addons..."
+        addons_location = SUNSTONE_LOCATION+"/public/app/addons/dist/*"
+        main_file_path = SUNSTONE_LOCATION+"/public/dist/main.js"
+        tmp_file = SUNSTONE_LOCATION+"/public/dist/main.tmp"
```
tinova: make them constants at the beginning of the file

src/sunstone/sunstone-server.rb
```diff
@@ -350,6 +350,60 @@ def destroy_session
...
+        tmp = File.open(tmp_file, "w")
+        main = File.new(main_file_path)
+        files = Dir[addons_location].select{ |f| File.file? f }
+        if files.length == 0 then return ""
+        end
+        l=main.gets
```
tinova: avoid using single letter variables

tinova: maybe use lines?

src/sunstone/sunstone-server.rb
```diff
@@ -350,6 +350,60 @@ def destroy_session
...
+                    add.each do |line|
+                        if line.include? "// list-start //"
+                            boolist = true
+                        end
+                        if !boolist
+                            tmp<<line
+                        end
+                    end
+                    add.close
+                end
```
tinova: create function for if body

src/sunstone/sunstone-server.rb
```diff
@@ -350,6 +350,60 @@ def destroy_session
...
+                            while line != nil
+                                tmp<<line
+                                line=add.gets
+                            end
+                        end
+                        line=add.gets
+                    end
+                    add.close
+                end
+            end
```
tinova: create function for if body

PR #1993, opened 2018-04-19
#1990 fixed bug when importing a template.

al3xhh:
bug applies to f1518 (merged in 1990)


Semedi:
So after this PR im able to import templates but it is not working for me speaking about wilds:

    $ onehost importvm 2 "_template00 - Cluster"

    There was an error trying to create a virtual network to repesent a vCenter network for a VM or VM 
    Template. Reason: undefined method `ipConfig' for nil:NilClass

It would be nice to check this behaviour in both environments (CLI, Sunstone)

PR #1996, opened 2018-04-19
vCenter driver is capable of import network names with slashes

Semedi:


src/vmm_mad/remotes/lib/vcenter_driver/network.rb
```diff
@@ -120,16 +120,19 @@ def self.to_one_template(opts = {})
         vm_or_template_name   = opts[:vm_or_template_name] || nil
         template_id           = opts[:template_id] || nil
 
+        bridge_name = network_name
+        network_name = network_name.gsub("/","_")
+
         network_import_name = generate_name(network_name, {:vcenter_name=>vcenter_instance_name, :dc_name=>dc_name})
 
-        one_tmp[:name]             = network_name
+        one_tmp[:name]             = bridge_name
```
tinova: shouldn't be using network name here?

Semedi: one_tmp is a temp hash that allow us to give feedback to the user in the time that he is importing the network but it is not used in the importation processs, so we can show either of two names.

is pointing to bridge_name due to is the real name

Issue #1692, opened 2018-01-29
Enable the option to add addons dynamically to Sunstone.


abelCoronado93:
# Enhancement Request
## Description
Enable the option to add addons dynamically to Sunstone.

## Use case
Add addons to Sunstone easily.

## Interface Changes
N/A

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

abelCoronado93:
- Needs documentation for new addons integration

PR #2067, opened 2018-05-07
F #1692: Enable the option to add addons dynamically to Sunstone

abelCoronado93:


src/sunstone/OpenNebulaAddons.rb
```diff
@@ -0,0 +1,89 @@
...
+require 'json'
+
+ADDONS_LOCATION = SUNSTONE_LOCATION + "/public/app/addons/dist/*"
+MAIN_DIST_PATH = SUNSTONE_LOCATION + "/public/dist/main-dist.js"
+MAIN_PATH = VAR_LOCATION + "/sunstone/main.js"
+
+class OpenNebulaAddons
+    def initialize(logger)
+        @logger = logger
+        @main_folder = File.join(VAR_LOCATION, "sunstone")
```
rsmontero: MAIN_PATH is constant but  main_folder is an instance variable, why?

src/sunstone/OpenNebulaAddons.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        begin
+            Dir.mkdir(@main_folder)
+        rescue Exception => e
+            @logger.error "Cannot create sunstone folder"
+            @logger.error e.message
+        end
+
+        main = File.open(MAIN_PATH, "w")
```
rsmontero: Should be new? also is better to set permissions for first time: File::CREAT|File::TRUNC|File::RDWR, 0644

src/sunstone/OpenNebulaAddons.rb
```diff
@@ -0,0 +1,89 @@
...
+        main_dist = File.new(MAIN_DIST_PATH)
```
rsmontero: Probably we could check if File.exsits?

src/sunstone/OpenNebulaAddons.rb
```diff
@@ -0,0 +1,89 @@
...
+
+            lines = main_dist.gets
+        end
+    end
+
+    private
+
+    def load_start_section(files, tmp)
+        files.each do |file|
+            add     = File.new(file)
```
rsmontero: Check if addon file exists

Issue #2084, opened 2018-05-11
vCenter Cache thread safe (thin related)


Semedi:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [ ] 5.4.4
- [ ] 5.4.5
- [ ] 5.4.6
- [ ] 5.4.7
- [ ] 5.4.8
- [ ] 5.4.9
- [ ] 5.4.10
- [ ] 5.4.11
- [x] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [ ] Storage & Images
- [ ] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [x] vCenter

## Description

vCenter cache is a static resource.

This can lead into some problems regarding thread data corruption with for example the thin server.
(This ruby server will spawn threads for Sunstone petition)

tinova:
probably not worth looking into, specially with the Sunstone rewrite.

PR #2086, opened 2018-05-11
B #2084: vCenter cache thread safety

Semedi:


rsmontero:
It's been decided to redesign the VIHelper class in a better way and thread-safe

Issue #1548, opened 2017-11-28
Time relative scheduled actions


tinova:
# Enhancement Request
## Description

This will enable interesting features such as automatic VM decomssion

## Use case

"I want this VM to expire automatically in 2 weeks"

## Interface Changes

Need to change SCHED_ACTIONS semantics. Add time relative dropdown, or time units, un SCHED ACTIONS tab in VM creation dialog in Sunstone

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

tinova:
Need also scheduled actions relative to the instantiation time.

PR #2097, opened 2018-05-16
F #1548:  Fixed bug with the empty action and the texts were changed when trying to create an instance of a vm

juanmont:
only master


juanmont:
already merged

PR #2128, opened 2018-05-28
Feature 2127

sergiojvg:
Add support for addons on the CLI


src/cli/command_parser.rb
```diff
@@ -1,5 +1,5 @@
 # -------------------------------------------------------------------------- #
-# Copyright 2002-2018, OpenNebula Project, OpenNebula Systems                #
+# Copyright 2002-2017, OpenNebula Project, OpenNebula Systems                #
```
tinova: year should be 2018

PR #2180, opened 2018-06-14
B #2173 always report RSS memory for KVM VMs

atodorov-storpool:
_max_memory_ is always equal to Available because:
  1. the memballoon is not used for resizing VM memory 
  2. the memballoon stats reporting is not enabled by default:

  ```
  <memballoon model='virtio'>
        <stats period='10'/>
    </memballoon>
  ```


xorel:
Fixed elsewhere https://github.com/OpenNebula/one/pull/2532

Issue #2238, opened 2018-06-29
main.js symlink not recovered after re-building sunstone js


atodorov-storpool:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [ ] 5.4.4
- [ ] 5.4.5
- [ ] 5.4.6
- [ ] 5.4.7
- [ ] 5.4.8
- [ ] 5.4.9
- [ ] 5.4.10
- [ ] 5.4.11
- [ ] 5.4.12
- [ ] 5.4.13
- [ ] 5.4.14
- [ ] 5.4.15
- [X] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [ ] Storage & Images
- [X] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## Description
When rebuilding the Sunstone GUI using /usr/lib//one/sunstone/public/build.sh the symlink from /usr/lib/one/sunstone/public/dist/main.js to /var/lib/one/sunstone/main.js destroyed leading to a "frozen" sunstone interface

### Expected Behavior
the symlink to /var/lib/one/sunstone/main.js should be recovered

### Actual Behavior
the symlink to /var/lib/one/sunstone/main.js is missing

## How to reproduce
```
cd /usr/lib/one/sunstone/public
build.sh -d
export PATH=$PATH:$PWD/node_modules/.bin
build.sh
```

A simple fix is to re-create the symlink after moving main.js to main-dist.js.

atodorov-storpool:
pull request https://github.com/OpenNebula/one/pull/2239

But not sure am I following the intended behavior...

abelCoronado93:
Let's choose your solution as an script option [-l] to recreate the links

atodorov-storpool:
Hi @abelCoronado93 ,

I've updated the workaround in our addon but forgot to update here.

Not sure `grunt sass` or `grunt-requirejs` is deleting the backed-up symlink `main.js-var`
[...]

PR #2250, opened 2018-07-03
B #2238: main.js symlink not recovered after re-building sunstone

abelCoronado93:


rsmontero:
Thanks @atodorov-storpool we prefer to keep the option

PR #2296, opened 2018-07-19
Feature 2291

al3xhh:
Don't squash both commits, they have been already squashed.


vholer:
Just a note -- this PR needs to be merged into master without squashing commits. And, only the da0b80d0c19a8b512486fa532ae43b145439638f should be cherry-picked into one-5.6.

Issue #2190, opened 2018-06-15
Onezone set should allow temporary zone changes


rsmontero:
# Enhancement Request
## Description

onezone set overwirte the .endpoint file, this makes the zone changes persistent and prevents the use of multiple sessions pointing to different zones.

## Use case
Use different zones from different shells

## Interface Changes
Add option to onezone set to not overwrite .endpoint

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2325, opened 2018-07-26
F #2190 Onezone set should allow temporary zone changes

joseangelgm:
cherrypick 5.6.1

### DIFF
--- a/src/cli/one_helper/onezone_helper.rb
+++ b/src/cli/one_helper/onezone_helper.rb
@@ -83,25 +83,30 @@ class OneZoneHelper < OpenNebulaHelper::OneHelper
         table
     end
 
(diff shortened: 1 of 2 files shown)


src/cli/one_helper/onezone_helper.rb
```diff
@@ -83,25 +83,30 @@ def format_pool(options)
         table
     end
 
-    def set_zone(zone_id)
-      zone = factory(zone_id)
-      rc = zone.info
+    def set_zone(zone_id, overwrite_file = true)
```
rsmontero: Prefer not to use options by default. Check that set_zone is not being used anywhere (this file should be only used in onezone command)

PR #2345, opened 2018-08-07
Update escaping

trenton42:
The Ruby library provides a method to escape `eq` queries: `equals`

https://github.com/ruby-ldap/ruby-net-ldap/blob/d78dc1ed51480f52117fd8be431de4fe8790b70f/lib/net/ldap/filter.rb#L148-L150

```ruby
    def equals(attribute, value)
      new(:eq, attribute, escape(value))
    end
```

Also change filter in `get_groups` from a string literal to a `Net::LDAP::Filter`


vholer:
Looks good, can be merged into master / one-5.6.

Issue #2356, opened 2018-08-20
Update Host hook triggers to include all possible states


OpenNebulaSupport:
**Description**
Host hook should include all possible states.

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

PR #2361, opened 2018-08-22
F #2356: more hook states for hosts

juanmont:


src/host/HostHook.cc
```diff
@@ -87,6 +87,14 @@ void merge_state(Host::HostState &state)
     {
         state = Host::DISABLED;
     }
+    else if (state == Host::MONITORING_MONITORED)
```
rsmontero: Change this for a swtich - case stanza so we do not miss another state update

PR #2364, opened 2018-08-23
B #2363 fixed IPAM free addr doesn't work.

al3xhh:
Cherrypick 5.6

### DIFF
--- a/src/vnm/AddressRangePool.cc
+++ b/src/vnm/AddressRangePool.cc
@@ -174,7 +174,6 @@ int AddressRangePool::from_xml_node(const xmlNodePtr node)
 }
 
 /* -------------------------------------------------------------------------- */
-/* -------------------------------------------------------------------------- */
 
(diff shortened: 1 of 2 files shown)


src/vnm/AddressRangePool.cc
```diff
@@ -196,6 +195,8 @@ int AddressRangePool::rm_ar(unsigned int ar_id, string& error_msg)
 
     AddressRange * ar_ptr = it->second;
 
+    ar_ptr->free_addr(ar_id, error_msg);
```
rsmontero: Why is this needed?  rm_ar only works on AR without addresses in use, so it would not be needed to free them.

Also free_addr uses index of an address (a IP for example)  not the ID of an address range.

al3xhh: That is needed to call the IPAM driver and remove the AR using the driver. In my case I'm developing the Packet IPAM driver, so without that, the AR is not delete in Packet, it is just delete in OpenNebula.

Issue #2140, opened 2018-05-31
vCenter importer: Wilds integration 


Semedi:
# Enhancement Request
## Description

Import wilds has the same splitted functionality like always, we need to integrate and reduce this code into the new importer.

## Use case

Import wilds from Opennebula Improve (Sunstone / CLI)

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

Semedi:
Also we have an issue with wild importation, Wilds machines cannot be removed completely there are more information here:

https://forum.opennebula.org/t/vcenter-remove-complete-wild-vm/6174

PR #2368, opened 2018-08-28
B #2140: reducing useless code on vcenter virtual machine (sunstone r…

Semedi:
…eferences)


tinova:
at last, sunstone is no more in the vcenter drivers :)

PR #2375, opened 2018-08-29
F #2356: Added new on states to oned.conf

juanmont:


juanmont:
cherry-pick to one-5.6

vholer:
Pls don't cherry-pick into `one-5.6`, avoid updating `oned.conf` only because of the comment.

PR #2390, opened 2018-09-05
Delete cs.po

p-bo:
Because there is cs_CZ, cs (which isn't used) alone is pointless. Please remove it from Transifex too. Thanks :)


abelCoronado93:
Done, removed from Transifex.

Issue #2320, opened 2018-07-26
Migrate images to/from KVM / vCenter DS


tinova:
**Description**

Automatic conversion to/from .vmdk to .raw/.qcow2/etc upon "oneimage clone" operations can be implemented

**Use case**

To facilitate smooth transitions between KVM and VMware based infrastructures, a

**Interface Changes**

None

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

tinova:
applied in 5.6 as well

tinova:
* qcow2 format attribute in image if target DS is qcow2
* check bus attribute
* add two 'guides' or 'sections'
   * from vmdk to qcow2/raw
   * from qcow2/raw to vmdk 
          * add note in documentation vmware tools

joseangelgm:
be able to change the driver and format is done.

joseangelgm:
PR test: https://github.com/OpenNebula/development/pull/403
PR code: https://github.com/OpenNebula/one/pull/2555

tinova:
Automatic conversion  .vmdk to .qcow2/.raw needs to be implemented

tinova:
doc reverted commit: 6cd72cbd24e6b81cee9f6af57819956257a5d653

PR #2396, opened 2018-09-06
F #2320 Migrate images to/from KVM / vCenter DS

joseangelgm:
Fix one bug and add if statement just in case the image and the origin have diferent image type.


joseangelgm:
Apply to 5.8 and 5.6

Issue #2428, opened 2018-09-18
Unable to update the Running Memory quota


abelCoronado93:
**Description**
Unable to update the Running Memory quota

**To Reproduce**
Try to update the Running Memory quota in User tab

**Expected behavior**
N/A

**Details**
 - Affected Component: Sunstone
 - Hypervisor: -
 - Version: 5.6.1

**Additional context**
N/A

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

PR #2430, opened 2018-09-18
B #2428: Unable to update the Running Memory quota

abelCoronado93:


abelCoronado93:
- master
- one-5.6-devel
- one-5.6

PR #2452, opened 2018-09-26
You mentioned you were having debates about fonts...

dabukalam:


tinova:
let's consider this for 6.0 .... (no!)

nice meeting you in Frankfurt

Issue #2437, opened 2018-09-19
vCenter problem with datastores [brackets and spaces] 


Semedi:
**Description**
template/vm importation could fail If you have a datastore with a name with spaces and/or brackets 

**To Reproduce**
Have vCenter with datastore (brackets/spaces in the name) imported in opennebula.

try to import any template/vm that belongs to that datastore.

**Expected behavior**

Import process ends successfully

**Details**
 - Affected Component: vCenter Driver
 - Hypervisor: vCenter
 - Version: development

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

PR #2463, opened 2018-10-02
B #2437: importing resources from vCenter ds with weird names

Semedi:
applies to Master


tinova:
applied also to one-5.6-devel

Issue #2439, opened 2018-09-19
Active disk snapshot is missing arrow in Sunstone


vholer:
**Description**
In the Sunstone, active disk snapshot is missing an arrow. There is just empty space.

![image](https://user-images.githubusercontent.com/414850/45766333-98591c00-bc37-11e8-87e8-35dcefed17bd.png)

**Details**
 - Affected Component: Sunstone
 - Version: 5.6.1 candidate

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

PR #2470, opened 2018-10-03
B #2439: Active disk snapshot is missing arrow in Sunstone

abelCoronado93:
- one-5.6
- one-5.6-devel
- master


tinova:
do we have tests for this?

abelCoronado93:
the problem was that an icon did not appear

tinova:
ok, we'll skip the test then

Issue #2474, opened 2018-10-04
Network model is not working in vCenter


OpenNebulaSupport:
**Description**
Network model is not working when using vCenter. The attribute is not taken from the template.

**To Reproduce**
Add a network model to the template and instantiate it.

**Details**
 - Affected Component: network
 - Hypervisor: vCenter

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

joseangelgm:
PR test https://github.com/OpenNebula/development/pull/381

tinova:
good work!

PR #2484, opened 2018-10-08
B #2474 Network model is not working in vCenter

joseangelgm:
Apply to:
- master
- one-5.6
- one-5.6-devel


tinova:
need to conform to ruby style

Issue #1147, opened 2017-11-20
Permit to rename disk snapshot names


OpenNebulaProject:
---

Author Name: **EOLE Team** (EOLE Team)
Original Redmine Issue: 4938, https://dev.opennebula.org/issues/4938
Original Date: 2016-11-23

---

Hello,

We heavily use the disk snapshots but sometimes the text of the name snapshot has a typo or should be changed to reflect the real state of the disk in that snapshot.

Could it be possible to make this text editable like for example the name of the instantiated VM.

Regards.

kvaps:
## Interface Changes

**XML-RPC API:**

 * add new `one.vm.snapshotrename` method with `string:<session>` `i4:<image_id>` `i4:<snap_id>`, `string:<new_name>` arguments

**Sunstone:**

 * Add new button: 🖉 -  for rename snapshots 

**Cli:**

*
    ```
    onevm snapshot-rename <vmid> <snapshot_id> <name>
    ```

PR #2485, opened 2018-10-08
F #1147: Permit to rename disk snapshot names in Sunstone

abelCoronado93:
- one-5.6-devel


tinova:
needs testing before merge

Issue #2491, opened 2018-10-09
install.sh doesn't accept -F and -g parameters


xorel:
**Description**
```
$ install.sh -h
Usage: install.sh [-u install_user] [-g install_group] [-k keep conf]
                  [-d ONE_LOCATION] [-c cli|ec2] [-r] [-h]
...
-G: install OpenNebula Gate
-f: install OpenNebula Flow
```

but when run with -f or -G it prints error:
```
install.sh: invalid option -- 'f'
install.sh: invalid option -- 'G'
```

**To Reproduce**
run: "install.sh -f"

**Expected behavior**
-f and -G parameters are accepted

**Details**
 - Affected Component: opennebula installation
 - Version: 5.6

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [x] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2492, opened 2018-10-09
B #2491: install.sh accepts -G and -f parameters

xorel:


tinova:
What branches does this go to? Suggestion, since it is a bug:

 * master
 * one-5.6
 * one-5.6-devel

xorel:
Agree with branches.
I also updated README.md.

Issue #2514, opened 2018-10-16
onedb crendentials for automation


jsfrerot:
**Description**
Allow onedb to load credentials from config file or env variable

**Use case**
When automating backup of opennebula, it would be useful to not have to hardcode the password

**Interface Changes**
onedb cli
maybe add option to load credentials from file or env variable. It could actually load the credentials from /etc/one/oned.conf

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
This issue has been already implemented in this [commit](https://github.com/OpenNebula/one/commit/12cc18d64b37584bb5fd3c1d472a3512a174435a).

So mainly the credentials are read from the `oned.conf` and you just need to use the command `onedb upgrade` for example, without any connection parameter.

Closing this issue.

PR #2516, opened 2018-10-17
F #2514 look for ONE_DB_PASSWORD environment var

atodorov-storpool:
before falling back to interactive


atodorov-storpool:
The addressed issue is still open...

PR #2517, opened 2018-10-17
B #2474 Network model is not working in vCenter

joseangelgm:


joseangelgm:
Apply to:
- master
- one-5.6
- one-5.6-devel

Issue #2438, opened 2018-09-19
TM ssh transfers inconsistent state if disk online


vholer:
**Description**
When TM ssh cpds action is triggered (e.g. via onevm disk-saveas), it can transfer inconsistent state of the disk via scp if VM is running and using the disk. There is no domain fsfreeze (or, suspend or better blockcopy) to ensure the image isn't modified on hypervisor during the transfer to image datastore.

https://github.com/OpenNebula/one/blob/d46535394667a037f21db67c5e2b4d9736e587bb/src/tm_mad/ssh/cpds#L49

(applies only for latest image state, not a snapshot)

**Details**
 - Affected Component: Storage
 - Hypervisor: KVM
 - Version: 4.14, 5.0, 5.2, 5.4, 5.6, master

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [ ] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2528, opened 2018-10-19
B #2438 Try blockcopy, fsfreeze or suspend first in ssh/cpds

xorel:


tinova:
looks good, is it possible to have a test? also, to which branches does it apply?

xorel:
apply for:
 * master
 * one-5.6

Added test for stopped VM (https://github.com/OpenNebula/development/pull/386)

Issue #2179, opened 2018-06-14
Real memory is reported only when it is more than the allocated memory (vmm_mad/kvm)


atodorov-storpool:
# Bug Report
## Version of OpenNebula

- [X] 5.2.2
- [X] 5.4.0
- [X] 5.4.1
- [X] 5.4.2
- [X] 5.4.3
- [X] 5.4.4
- [X] 5.4.5
- [X] 5.4.6
- [X] 5.4.7
- [X] 5.4.8
- [X] 5.4.9
- [X] 5.4.10
- [X] 5.4.11
- [X] 5.5.80
- [X] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [X] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [ ] Storage & Images
- [ ] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## Description
In current configuration `virh dominfo one-$vmid` always returns the allocated memory. The second source of information is the RSS from the `ps` output which is the real memory usage by the guest on the host (It could be bigger because it is the sum of the qemu-kvm process itself and the domain usage).

I think that the RSS value is relevant in this case. 

### Expected Behavior
The monitoring (vmm/kvm/poll) should return the real memory usage (RSS)

### Actual Behavior
The monitoring (vmm/kvm/poll) returns the value from _dominfo_ or _ps-RSS_ depending on which is bigger (_MAX(dominfo,rss)_)

## How to reproduce
1. deploy a Linux VM with relatively large RAM (8GB)
2. look the reported stats (onevm or sunstone)

atodorov-storpool:
https://github.com/OpenNebula/one/pull/2180

atodorov-storpool:
The issue was re-introduced with commit https://github.com/OpenNebula/one/commit/8ba51f0933894b88ffcf8557bc62d9400cf0846d

vholer:
I think it's just poor conflicts handling. It was fixed back again in
e3969fd9d68bb0240ccca119a04c288c49561c3f.

PR #2532, opened 2018-10-23
B #2179 Report RSS + swap usage for KVM VMs

xorel:


xorel:
Apply for:
 * one-5.6
 * one-5.6-devel
 * master

Issue #2161, opened 2018-06-07
Support for VXLAN with BGP EVPN  / MP-BGP


smangelkramer:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [ ] 5.4.4
- [ ] 5.4.5
- [ ] 5.4.6
- [ ] 5.4.7
- [ ] 5.4.8
- [ ] 5.4.9
- [ ] 5.4.10
- [ ] 5.4.11
- [ ] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [ ] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [ ] Storage & Images
- [ ] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## How to reproduce

# Enhancement Request
## Description

From my point of view a modern datacenter fabric has to scale
and if based on VXLAN a control plane is one of the secrets to
gain scalability and manageability for a resilent and performant
multi-tenant datacenter and/or cloud network.

Therefore i did some PoC for adding VXLAN with BGP-EVPN support to OpenNebula.
For sure, OpenNebula is the ideal platform to extend and integrate with SDN and other tools (pre,post,clean).

My PoC (now also running in a medium sized cloud environment / datacenter is based on

- VXLAN  (RFC7348)
- BGP4 (RFC4271)
- BGP MPLS-Based Ethernet-VPN (RFC 7432)
- draft-ietf-bess-evpn-overlay-08 
- MP-BGP (RFC4760)

The advantages over VXLAN with Multicast are:

- traffic engineering
- more scalability
- better manageability
- easier to handle than IP-Multicast in a large scale network

In my setup BGP EVPN and the whole VXLAN is based on classic Linux bridging (one Bridge per VNI)
and Quagga (Cumulusnetworks) inside a Docker container on the VMM hosts.

On the (mostly redundant) OpenNebula frontends there are BGP Route Reflectors installed (also dockerized Quagga).

Example for the RR:

```
docker pull cumulusnetworks/quagga
[...]

PR #2542, opened 2018-10-25
F #2161 Support 2 modes for vxlan multicast and EVPN

xorel:


xorel:
apply for:
 * master
 * one-5.6-devel
 * one-5.6

PR #2545, opened 2018-10-25
Add support for linuxcontainers addon

sergiojvg:


share/pkgs/Ubuntu/opennebula.sudoers
```diff
@@ -9,5 +9,8 @@ Cmnd_Alias ONE_OVS = /usr/bin/ovs-ofctl, /usr/bin/ovs-vsctl
 Cmnd_Alias ONE_XEN = /usr/sbin/xentop, /usr/sbin/xl, /usr/sbin/xm
 Cmnd_Alias ONE_CEPH = /usr/bin/rbd
 Cmnd_Alias ONE_HA = /bin/systemctl start opennebula-flow, /bin/systemctl stop opennebula-flow, /bin/systemctl start opennebula-gate, /bin/systemctl stop opennebula-gate, /usr/sbin/service opennebula-flow start, /usr/sbin/service opennebula-flow stop, /usr/sbin/service opennebula-gate start, /usr/sbin/service opennebula-gate stop
+Cmnd_Alias ONE_MOUNT = /bin/mount /var/tmp/*.raw /var/tmp/*, /bin/umount /var/tmp/*
```
rsmontero: This should be added to https://github.com/OpenNebula/one/blob/master/share/sudoers/sudo_commands.rb

Also review use of wild char * in sudo

vholer: Also, all platforms should be refreshed. Ideally, all these commands should be better under a single alias, so they can be easily identified and disabled (e.g., ONE_LXD or LXC).

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -0,0 +1,161 @@
...
+    ;;
+*)
+    terminal="/bin/bash"
+    read -r -d '' commands << EOT
+        echo "This distro is not supported by OpenNebula context" > /root/opennebula_context.log
+EOT
+    ;;
+esac
+
+cat << EOF | sudo chroot $tmp_dir/$id $terminal
```
vholer: Can't be used `fakeroot` somehow? Enable sudo chroot is big security issue.

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -0,0 +1,161 @@
...
+        mkfs.xfs -f $tmp_dir/$id.raw > /dev/null 2>&1
+        ;;
+    *)
+        mkfs.ext4 -F $tmp_dir/$id.raw > /dev/null 2>&1
+        ;;
+esac
+
+
+mkdir $tmp_dir/$id
+sudo mount $tmp_dir/$id.raw $tmp_dir/$id
```
vholer: I think there could be a different way if we focus only on ext4. Do all the necessary changes in the chroot and then copy into ext4 raw image directory by directory and file by file via `debugfs` command. Haven't tried it, it's a much harder way, but like it more than mount/sudo chroot... https://www.cs.montana.edu/courses/309/topics/4-disks/debugfs_example.html

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -0,0 +1,161 @@
...
+        $curl $context_url/v$selected_tag/one-context-$selected_tag-1.el7.noarch.rpm -Lsfo /root/context.rpm
+        yum install /root/context.rpm -y > /dev/null 2>&1
+EOT
+    ;;
+*alpine*)
+    terminal="/bin/bash"
+    read -r -d '' commands << EOT
+        echo "nameserver $dns_server" > /etc/resolv.conf
+        $curl $context_url/v$selected_tag/one-context-$selected_tag-1.el7.noarch.rpm -Lsfo /root/context.rpm
+        yum install /root/context.rpm -y > /dev/null 2>&1
```
vholer: Alpine is `apk add --allow-untrusted ...`

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -0,0 +1,161 @@
...
+    terminal="/bin/bash"
+    read -r -d '' commands << EOT
+        echo "This distro is not supported by OpenNebula context" > /root/opennebula_context.log
+EOT
+    ;;
+esac
+
+cat << EOF | sudo chroot $tmp_dir/$id $terminal
+$commands
+EOF
```
vholer: There should be some cleanup after `$commands` shared across all the appliances. E.g.:

- `rm -f /etc/resolv.conf`
- `rm -rf /etc/ssh/ssh_host_*`
- `usermod -p '*' root`

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -0,0 +1,161 @@
+#!/bin/bash
+
+get_tag_name () {
+    version=`oned -v | grep "is distributed" | awk '{print $2}'`
```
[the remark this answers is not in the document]
sergiojvg: Not sure if it's better to depend on a file that could even be deleted instead of the stable output of a command that will always succeed. @rsmontero, what do you think? Although...if we use that file to get the version number at the frontend for other tasks it could be better to unify.

src/datastore_mad/remotes/lxd_downloader.sh
sergiojvg: Haven't tried debugfs before, but, as you said, it only supports ext, and it is a requirement for us to support more filesystems at the moment (like XFS). I tried several FUSE mount tools, having the best results with guestmount but it  was not good: dependencies issues, a daemon (forgot which) must be configured and started, there was even a bug on ubuntu 18.04 that needs to be fixed before using it...so I went back to sudo mount, too many things could go wrong.

src/datastore_mad/remotes/lxd_downloader.sh
sergiojvg: Thanks! Fixed.

src/datastore_mad/remotes/lxd_downloader.sh
sergiojvg: Great idea

vholer:
I was thinking more about it and the least dirty solution looks to me to move the appliance patching functionality into a separate script, which will be called from the downloader, ship this script as part of the main OpenNebula code (not remotes), and allow sudo to this script.

I.e., have `/usr/share/one/tar2img` with enabled sudo to this scripts. Arguments could be

1. input tar archive
2. output image
3. image type
4. image size
5. filesystem

... and **on STDIN pass all the commands to apply in the image**.

Example:
```
$ /usr/share/one/tar2img c7.tar.xz c7.img qcow2 2048 xfs <<EOF
        yum install /root/context.rpm -y > /dev/null 2>&1
        echo "#This file is modified by OpenNebula. Don't write in here" > /etc/resolv.conf
        rm -f /etc/ssh/ssh_host_* > /dev/null 2>&1
        usermod -p '*' root > /dev/null 2>&1
EOF
```

src/datastore_mad/remotes/lxd_downloader.sh
sergiojvg: Tried but couldn't. Containers from linuxcontainers have root as owner and group for most files insode .tar.xz. When uncompressing, I need to respect ownership (I can't simply massively change ownership to everything). Because everything belongs to root, fakeroot won't work.

bmullan:
Just butting in but I'm anxious to try this when its ready.

Perhaps ask Stephane Graber (the LXD project lead) if he has a solution if
needed?

his email:  stgraber@ubuntu.com

Brian Mullan
Raleigh, North Carolina USA

PR #2553, opened 2018-10-26
 F #1745: Admin should be able to disable Advanced options from Cloud view

al3xhh:


tinova:
committed to one-5.6-devel as well

PR #2569, opened 2018-10-31
Doing auto refresh only in info tab

juanmont:
cherry pick to one-5.6


Madko:
is the info tab where you choose owner and group ? Because refresh prevent us to do that with 5.6.2

Issue #911, opened 2017-11-20
IP(v6) alias(es) support


OpenNebulaProject:
---

Author Name: **Stefan Kooman** (Stefan Kooman)
Original Redmine Issue: 3784, https://dev.opennebula.org/issues/3784
Original Date: 2015-04-30

---

This feature has been discussed before but somehow never made it to a feature request: http://lists.opennebula.org/pipermail/users-opennebula.org/2014-January/026313.html. The idea is to have the possibility to add IP(v6) alias(es) to a TEMPLATE and / or a running VM. This way it's possible to (still) have all addresses / leases managed by OpenNebula. Normally a "true" IP alias would get a subnet mask of "/32" (IPv4) or "/128" (IPv6). If the IP(v6) alias inherits the properties from the VNET / AR this subnet mask will not be correct. Ideally there would be an option to override the properties of the VNET / AR. To make this IP(v6) alias(es) work when "IP address hijack prevention" is active (openvswitch) a new OpenFlow rule needs to be added for every IP(v6) alias added (related issue #3181). If an alias would be added to a running VM a driver should be executed on the host the VM is running on to add this OpenFlow rule. Context scripts could be enhanced to automatically add the IP(v6) alias(es) (ip (-6) addr add ip/subnetmask dev interface). If it would be possible to override the mac-address in an AR reservation this could also solve the (virtual) ip-address / MAC (VRRP / CARP) problem.

OpenNebulaProject:
---

Original Redmine Comment
Author Name: **Ruben S. Montero** (@rsmontero)
Original Date: 2015-05-13T16:17:37Z

---

After some discussions in the forum, we've come out with the following design

```
NIC = [ NETWORK="public" NAME="main" ]
NIC = [ NETWORK="public" ALIAS="main" ]

```

Interface with ALIAS attribute will not be included in the deployment file for the hypervisor. But it would be included in context as an alias for the target NIC.

OpenNebulaProject:
---

Original Redmine Comment
Author Name: **Stefan Kooman** (Stefan Kooman)
Original Date: 2015-10-16T08:05:17Z

---

This feature does not (completely) solve the "(virtual) ip-address / MAC (VRRP / CARP)" problem. As it is currently not possible to request more than 1 "LEASE" from OpenNebula in a given AR / VNET. A special IP(6) could be introduced that has the ability to have > 1 LEASE (allow duplicates) and as such the VIP(6) be used in more than 1 VM.

[...]

PR #2628, opened 2018-11-21
F #911: IP(v6) alias(es) support

al3xhh:
* Modify VM context generation.
   * Modify libvirt deployment file generation.
   * Modify detach nic functionality.
   * Modify atach nic to not call the driver.
   * Modify CLI nic-attach, so an alias can be attached.


hydro-b:
Great to see this landing in ONE!

PR #2647, opened 2018-11-26
[vcenter] instantiate as persistent, reduce overhead, starting bootOrder

Semedi:
**NO SQUASH**

### DIFF
--- a/src/rm/RequestManagerClone.cc
+++ b/src/rm/RequestManagerClone.cc
@@ -184,7 +184,7 @@ Request::ErrorCode VMTemplateClone::clone(int source_id, const string &name,
             oss << name << "-disk-" << ndisk;
 
             ec = img_clone.request_execute(img_id, oss.str(), -1,
-                    (*disk)->is_managed(), new_img_id, img_att);
+                    true, new_img_id, img_att);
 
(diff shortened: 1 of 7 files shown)


Semedi:
only master

src/rm/RequestManagerClone.cc
```diff
@@ -184,7 +184,7 @@ Request::ErrorCode VMTemplateClone::clone(int source_id, const string &name,
             oss << name << "-disk-" << ndisk;
 
             ec = img_clone.request_execute(img_id, oss.str(), -1,
-                    (*disk)->is_managed(), new_img_id, img_att);
+                    true, new_img_id, img_att);
```
tinova: maybe we don't need this parameter anymore @rsmontero ?

tinova:
lgtm but ideally would want to have @rsmontero review the core change

src/rm/RequestManagerClone.cc
Semedi: i think there is another call using the parameter, thats why im not removing it.

Semedi:
need to talk with @rsmontero before the merge.

PR #2655, opened 2018-11-29
Feature 1684 fix2

dann1:
Fix bugs detected with microenvs testing

### DIFF
--- a/src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
+++ b/src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
@@ -368,7 +368,7 @@ class Mapper
                     partitions = [partitions]
                 end
 
-                partitions.delete_if { |p|  
+                partitions.delete_if { |p|
                     p['fstype'].casecmp?('swap') if p['fstype']
                 }
             end
--- a/src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
+++ b/src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
@@ -66,7 +66,9 @@ class Mapper
         :e2fsck     => 'sudo e2fsck',
         :resize2fs  => 'sudo resize2fs',
         :xfs_growfs => 'sudo xfs_growfs',
-        :chmod_nfs  => 'chmod o+w'
+        :chmod_nfs  => 'chmod o+w',
+        :rbd        => 'sudo rbd --id',
+        :df         => 'sudo df -h'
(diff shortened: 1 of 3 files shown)


src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
```diff
@@ -182,15 +184,23 @@ def unmap(one_vm, disk, directory)
...
 
         sys_parts  = lsblk('')
-        partitions = []
-        device     = ''
 
         return false if !sys_parts
 
+        partitions = []
+
+        df = "#{COMMANDS[:df]} #{directory} | grep /dev"
```
rsmontero: Not sure if this work as execute works directly through popen3, there is no shell setting up the | to grep

src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
```diff
@@ -182,15 +184,23 @@ def unmap(one_vm, disk, directory)
...
+        rc, device, err = Command.execute(df, false)
```
rsmontero: Parse stdout through each_line and ~= or .match operators

src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
```diff
@@ -182,15 +184,23 @@ def unmap(one_vm, disk, directory)
...
+        device = device.chomp.split(' ')[0]
+
+        if rc != 0
+            OpenNebula.log_error("unmap: #{err}")
+            return
+        end
+
         sys_parts.each { |d|
-            if d['mountpoint'] == directory
+            if d['path'] == device
```
rsmontero: Why is this needed, does not lsblk output include the mountpoint? At this point it should (because of at unmap action) When it is not included

src/vmm_mad/remotes/lib/lxd/mapper/qcow2.rb
```diff
@@ -28,25 +28,19 @@ class Qcow2Mapper <  Mapper
...
-
+        rc, _out, err = Command.execute(cmd, true)
+        
         if rc != 0 
             OpenNebula.log_error("do_map: #{err}")
-            return nil
+            return
         end
+        
+        sleep 0.5 # TODO: improve settledown, lsblk -f fails
```
rsmontero: Why the check with the lsblk to check fstype has been remove , lsblk should not fail

Issue #2656, opened 2018-11-29
[GOCA] Implement missing XML RPCs


jp-pro:
**Description**
Goca does not support all XML RPCs. It should be nice to have all call available in GO.

**Use case**
Having all XML RPC available in GO can help developpers doing tools in go.

**Interface Changes**
No change expected on OpenNebula as this is only adding bindings in the Goca.

**Additional Context**
I am adding missing calls. I will create a pull request with my suggestion

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2659, opened 2018-11-29
F #2656: [GOCA] add missing XML RPC Go Bindings

jp-pro:
This pull request implements missing XML RPC in Goca:
* security group
* marketplace
* marketplace app
* virtual router
* system config

### DIFF
--- a/src/oca/go/src/goca/marketplaceapp.go
+++ b/src/oca/go/src/goca/marketplaceapp.go
@@ -71,7 +71,7 @@ func NewMarketPlaceAppFromName(name string) (*MarketPlaceApp, error) {
 // CreateMarketPlaceApp allocates a new marketplace app. It returns the new marketplace app ID.
 // * tpl: template of the marketplace app
 // * market: market place ID
-func CreateMarketPlaceApp(tpl string, int market) (uint, error) {
+func CreateMarketPlaceApp(tpl string, market int) (uint, error) {
 	response, err := client.Call("one.marketapp.allocate", tpl, market)
 	if err != nil {
 		return 0, err
--- a/src/oca/go/src/goca/marketplaceapp.go
+++ b/src/oca/go/src/goca/marketplaceapp.go
@@ -1,0 +1,172 @@
+package goca
+
+import "errors"
+
+// MarketPlaceApp represents an OpenNebula MarketPlaceApp
+type MarketPlaceApp struct {
+	XMLResource
+	ID   uint
+	Name string
+}
+
+// MarketPlaceAppPool represents an OpenNebula MarketPlaceAppPool
+type MarketPlaceAppPool struct {
+	XMLResource
+}
+
+// NewMarketPlaceAppPool returns a marketplace app pool. A connection to OpenNebula is
+// performed.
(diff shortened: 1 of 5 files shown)


src/oca/go/src/goca/marketplaceapp.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return NewMarketPlaceApp(id), nil
+}
+
+// CreateMarketPlaceApp allocates a new marketplace app. It returns the new marketplace app ID.
+// * tpl: template of the marketplace app
+func CreateMarketPlaceApp(tpl string) (uint, error) {
```
christian7007: It's needed the ID of the destination marketplace in which the app will be store.
```suggestion
func CreateMarketPlaceApp(tpl string, market int) (uint, error) {
```

src/oca/go/src/goca/virtualrouter.go
```diff
@@ -0,0 +1,183 @@
...
+	return err
+}
+
+// Instantiate will instantiate the virtual router. It returns the ID of the new VM
+// * number: Number of VMs to instantiate.
+// * tplid: VM Template id to instantiate.
+// * name: Name for the VM instances. If it is an empty string OpenNebula will set a default name. Wildcard %i can be used.
+// * pending: False to create the VM on pending (default), True to create it on hold.
+// * extra: A string containing an extra template to be merged with the one being instantiated. It can be empty. Syntax can be the usual attribute=value or XML.
+func (vr *VirtualRouter) Instantiate(number, tplid int, name string, pending bool, extra string) (uint, error) {
```
christian7007: I think it would be better to rename the pending parameter by hold. It seem weird to me to call it pending when it need to be false for the VM to be instantiated in pending state.
```suggestion
func (vr *VirtualRouter) Instantiate(number, tplid int, name string, hold bool, extra string) (uint, error) {
```

jp-pro: indeed!

src/oca/go/src/goca/marketplaceapp.go
jp-pro: I agree. In that case we need to use the market in the client call too ;).
In the doc it is written: "IN | Int | The datastore ID." while it is the marketAPP id. That was wrong for me.

jp-pro: I just fixed it

christian7007: Yes, you need to pass the parameter to the call below.
Thanks for notice about the docs, I'll fix that.

christian7007: Here you misplaced the type "int market" it would be "market int"

jp-pro: Oh Indeed. I fixed that

src/oca/go/src/goca/virtualrouter.go
jp-pro: yes. fixed in another commit

PR #2661, opened 2018-11-29
B #2658: GOCA - fix Info methods

treywelsh:
This pull request fix the Info method of the host in host.go file from go bindings


treywelsh:
I just see that other following files were also affected by the same error :
cluster.go datastore.go group.go user.go vdc.go virtualnetwork.go zone.go

Do I need to do an other ticket with pull request ? or can I update this one ?
I'm sorry !

christian7007:
Don't worry, you can just update the title of the issue and the PR and keep pushing the changes here.

Issue #2662, opened 2018-11-29
[GOCA] add state values and strings for the host


treywelsh:
**Description**
In the file host.go from the go bindings, there is no values and associated strings for the State of the host.
It's done for other elements, see other files image.go or vm.go.
Theses values are documented in the XMLRPC API documentation.

**Use case**
I should be more easy to develop tools with it.

**Interface Changes**
It will not change the existing interface, it's about adding an enum of values an associated String method.

**Additional Context**
I will create a pull request with my suggestion

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2663, opened 2018-11-29
F #2662: GOCA - Add host state values and strings

treywelsh:
This pull request add values and string for the state of the hosts in goca
It's documented in the XMLRPC api and I just propose to add corresponding the code.
It's implemented for others elements as VM, images ...


treywelsh:
The automatic review had failed, so I just updated my branch to meet the requirements

PR #2670, opened 2018-12-01
Feature 2650

sergiojvg:


src/vmm_mad/remotes/kvm/poll
```diff
@@ -157,8 +158,7 @@ module KVM
             end
 
             if (values[:state] == 'a') && (vm[:reason] != 'migrating')
-                values.merge!(get_interface_statistics(name, xml))
-                values.merge!(get_diskio_statistics(name, xml))
+                values.merge!(get_io_statistics(name, xml))
```
vholer: 1. Similar change should be done in method `def self.get_vm_info(one_vm)`.
2. Methods `get_intertface_statics` and `get_diskio_statistics should` should be removed as they provide redundant functionality.

src/vmm_mad/remotes/kvm/poll
```diff
@@ -396,6 +396,30 @@ module KVM
         end
     end
 
+    def self.get_io_statistics(vmid, text = nil)
+        #disks=get_disk_names(vmid, text)
```
vholer: Forgotten comments, here and below.

src/vmm_mad/remotes/kvm/poll
```diff
@@ -396,6 +396,30 @@ module KVM
...
+        #if disks && !disks.empty?
+        values={}
+        values[:netrx] = 0
+        values[:nettx] = 0
+        values[:diskrdbytes]=0
+        values[:diskwrbytes]=0
+        values[:diskrdiops]=0
+        values[:diskwriops]=0
+
+        vm_stats.scan(/rx.bytes=(\d*)/).each { |value| values[:netrx]+=value[0].to_i }
```
vholer: If this is all is for the performance reasons, it would be better to have just a single pass over the `vm_stats` (similar way as old methods `get_diskio_statistics` etc.).

Also, regular expression better with at least one number `(\d+)`.

src/vmm_mad/remotes/kvm/poll
```diff
@@ -51,6 +51,7 @@ module KVM
         :dumpxml    => 'virsh --connect LIBVIRT_URI --readonly dumpxml',
         :domifstat  => 'virsh --connect LIBVIRT_URI --readonly domifstat',
```
vholer: `:domifstat` and `:domblkstat` are now probably obsolete (???)

Issue #2678, opened 2018-12-04
[GOCA] fix and modify error management of client


treywelsh:
**Description**

It concerns the client error management in the goca.go file of the Go bindings and there is two commits on this branch:

**The first commit** remove an error check :
The function SetClient should return an error, but just return a "nil" value.
An other part of the code check for this error, but then check a "nil" value.

First of all, I remove this part because it's useless, and the error is checked later in "Call" function line 127.

It also changes the way the error management was initially thought :
the check is made in the "init" function of the package and call log.Fatal in case of an error.
The current way of handling errors with the xmlrpc client is : if it fails, panic.

So, fixing the issue in this way of thought (error = panic) would be to return an error instead of "nil" in SetClient function, causing the goca package to panic in case of an error (at the package import because of the "init" function).
 
Why I fix this bug in this way ?
It allows the user to manage the error (when the "Call" function fail), and maybe to call again SetClient function. 

This approach to fix the bug follow another error handling strategy: return better errors and messages and remove all the calls allowing the library to stop the main program.
I need to use goca as a little part of a more global tool. If goca fail, the whole program may fail (or I need to use recover in the main tool to patch this behaviour).

**The second commit** remove remaining log.Fatal calls and just return specialized errors instead.
It allow to use type assertion to better mana errors.

Futher work:
To improve this I plan to parse the full XML response (add error codes) of OpenNebula soon (actually, the error code is just not used) and return it to the user to let him handle this.

**To Reproduce**
First commit: Reproduce each time.
Second commit: Reproduce at each client error.

**Expected behavior**
First commit: change nothing in the behaviour, because of an error.
Second commit: don't panic anymore on xmlrpc client error.

**Details**
 - Affected Component: goca client part
 - Version: development (master)

**Additional context**
I will provide a fix in a pull request

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

treywelsh:
All my apologizes I made an error it, the goca part on master branch doesn't compile anymore, I propose a commit to fix it.
I wasn't compiling the right thing, It compiles now.

Sorry

treywelsh:
In this last pull request, I add the parsing of the error code from the opennebula reponse and return it to the user.
It allow a fine grained error management

[...]

PR #2679, opened 2018-12-04
B #2678: GOCA - fix and modify xmlrpc client error checking

treywelsh:
I submit a branch with two commit:
The first providing a way to fix an error
The second suggesting to replace remaining fatal calls with specialized errors

The main effect is to allow errors handling in a better way with type assertion, and to avoid goca to kill the main program


treywelsh:
I update my pull request to improve it and fix a little error

treywelsh:
I just updated my pull request to improve it, and fix an error.

PR #2729, opened 2018-12-16
LXD fixes

dann1:
- Improved/fixed logging
- Fixed storage issues
- Fixed reverted BASE_PATH detection


src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb
```diff
@@ -185,40 +185,42 @@ def unmap(one_vm, disk, directory)
         sys_parts = lsblk('')
 
         return false unless sys_parts
- 
+
         partitions = []
         device = ''
-        real_path = directory
+
+        dir_ok = directory.dup.gsub('//', '/')
```
rsmontero: Why is this needed? Double // should be ok in a path. Does it break any addtional parsing?

dann1: > Why is this needed? Double // should be ok in a path. Does it break any addtional parsing?

because of [this](https://github.com/dann1/one/blob/e55c247c24cff72b0bd054e3be1d2be01457c8ed/src/vmm_mad/remotes/lib/lxd/mapper/mapper.rb#L202)

Issue #2677, opened 2018-12-04
LDAP: Case sensitive group-membership matching


bjozet:
**Description**
When using LDAP login, `src/authm_mad/remotes/ldap/ldap_auth.rb` checks if the user is member of mapped groups (in `get_groups`-function). These mapped groups gets compared with case-sensitivity, although many/most/all(?) LDAP DNs are case-insensitive, leading to some frustration for an admin if LDAP objects change or if you typo:ed the mapping with wrong case in template or file, which breaks login for LDAP users.

**To Reproduce**
Have a ldap user which belongs to group `CN=Donkey,DC=Kong`, and try to log in, when having a map in mapping_filename (either manually created or through TEMPLATE) which contains a map of `cn=Donkey,dc=Kong` to a group id.

**Expected behavior**
Group comparison should not care for case, if LDAP is case-insensitive.

**Details**
 - Affected Component: Sunstone
 - Version: development

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2732, opened 2018-12-17
B #2677 Match LDAP groups case insensitive

xorel:


xorel:
Test update: https://github.com/OpenNebula/development/pull/451

Issue #2652, opened 2018-11-28
[vCenter] Boot order functionality


Semedi:
**Description**

Capability of use BOOT attribute of OpenNebula in vCenter.

Allow you to choose the boot order for the vm.

**Use case**

Import a vCenter template, and change the boot order of the spawned vms.

**Interface Changes**

We need to add the same section of OS | BOOT that is available in KVM sunstone but not in vCenter view.

**Additional Context**

The mandatory thing is to choose between disk order. 
This feature aims to provide that functionality, we will consider to keep this feature until the whole process boot is done (nics and other kind of devices) or make a new one.

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

Semedi:
For the moment is only going to work with unmanaged disks (safe feature), however is going to be possible to use any kind of disk soon.

PR #2736, opened 2018-12-17
F #2652 [vCenter] new nic model and its possible to choose nic bootorder

Semedi:


src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -128,6 +128,28 @@ class Nic < Resource
         def initialize(id, one_res, vc_res)
             super(id, one_res, vc_res)
         end
+
+        def self.one_nic(id, one_res)
+            @error_message = "vCenter device does not exist at the moment"
```
tinova: would it be better to have a @nic_type as a class variable, instead of using @error_message?

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -128,6 +128,28 @@ class Nic < Resource
...
+
+            self.new(id, one_res, nil)
+        end
+
+        def self.vc_nic(vc_res)
+            @error_message = "one nic does not exist at the moment"
```
tinova: It's better to use OpenNebula instead of "one"

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -1092,6 +1200,32 @@ def set_boot_order(boot_info)
         RbVmomi::VIM.VirtualMachineBootOptions({bootOrder: boot_order})
     end
 
+    def sync_nics(option = :none, execute = true)
```
tinova: Add a description of what this does. Are we removing all NICs? Only removing the ones missing from OpenNebula? A short, concise paragraph suffices.

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -1092,6 +1200,32 @@ def set_boot_order(boot_info)
...
+
+        spec_hash = { :deviceChange => device_change }
+
+        spec = RbVmomi::VIM.VirtualMachineConfigSpec(spec_hash)
+        @item.ReconfigVM_Task(:spec => spec).wait_for_completion
+
+        info_nics
+    end
+
     # TODO
```
tinova: is the TODO needed?

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -1092,6 +1200,32 @@ def set_boot_order(boot_info)
...
     # Synchronize the OpenNebula VM representation with vCenter VM
```
tinova: Briefly explain what this does with NICs and DISKs, if they a) exist in both vcenter and ONE, b) exist only in vcenter, c) exist only in ONE

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -1303,6 +1438,8 @@ def calculate_add_nic_spec(nic)
                  :port => port)
         end
 
+        # grab the last unitNumber to ensure the nic to be added at the end
+        @unic = @unic || get_vcenter_nics.map{|d| d.unitNumber}.max rescue 0
```
tinova: maybe we need this to fail with an error message instead of 'rescue 0'?

Semedi: rescue 0 means that get_vcenter_nics array is empty, so we are rescuing this with the 0 unitNumber

src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
Semedi: @error message is only an error to be displayed when u ask for a vcenter_device having only the one xml, thats why we only have @error_message in one_nic and one_disk.

Issue #2756, opened 2018-12-21
[GOCA] fix tests


treywelsh:
**Description**
1) The tests does not build because of trying to import the goca package (the tests are already in the goca package, the import is useless)

2) In case of failure when creating a new entity (for instance in calling NewTemplateFromName), the tests panics because of an error in the "Info" method called just below.
It's because we call "Info" method on a nil pointer (we access template.ID field inside).

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x13b061b]
```

The whole test suite ends here. (for all *_test.go files).
The problem is: we may want the other tests from other *_test.go files to continue.
In replacing some Error calls by Fatal calls, the "execution will continue at the next test or benchmark"
The current XXX_test.go file end here, and we go to the next.

**To Reproduce**
Set a bad URL for xmlrpc imply that all test will fail and the error will occur.

**Expected behavior**
If a test fail, I don't want all tests to fail.
In addition, should not call a method on a nil pointer, the user should check this.

**Details**
 - Affected Component: GOCA
 - Version: development (master branch)

**Additional context**
I will create a pull request with my fix

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #2757, opened 2018-12-21
 B #2756: GOCA - fix tests

treywelsh:
Fix GOCA tests: fix compilation and fix unexpected behavior


treywelsh:
The review does not pass but it's due to previous code. 
The goal of this pull request is not to fix this.

Issue #1684, opened 2018-01-25
Support for lxd containers


tinova:
# Enhancement Request
## Description

lxd provides a number of advantages, mainly performance-wise, over traditional VMs

## Use case

This feature proposes the management of lxd as first class citizens (same as VMs are managed by OpenNebula now).

## Interface Changes

This is a transversal features that implies changes and additions in all OpenNebula layers.

# Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [x] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

atodorov-storpool:
I am using Ubuntu 18.04 VM from the MarketPlace as LxD host. and the monitoring of the container is failing because the path to the memory cgroup differ '/sys/fs/cgroup/memory/sysdefault/lxc/' vs the hard-coded '/sys/fs/cgroup/memory/lxc/'. I think that the path to the cgroup should be configurable via configuration variable...

PR #2779, opened 2019-01-04
F #1684 add LXD support for disk TYPE=BLOCK

atodorov-storpool:
* add '-s' to the _file_ command to read special devices
* add '-s' to enable "sync mode" _kpartx_ when adding partition mappings

and handle the DISK case for disk['TYPE']


atodorov-storpool:
will rewrite the commit message

Issue #2830, opened 2019-01-18
[GOCA] migrate xml package from xmlpath to encoding/xml


treywelsh:
**Description**
The goal of this migration is to use the encoding/xml package instead of xmlpath in order to decode the XML content from OpenNebula.
For each OpenNebula entity, a set of structures will be defined to which is directly mapped the XML content.
The developper will work directly with theses structures.
Sometimes due to dynamic parts he will have to do some tests but less than with xmlpath module.

The types conversions and parsing part are removed from the developper code,
in this way, his code will be much more simple, and this also allow auto completion.

**Use case**
It's useful for every XXX.info requests.

**Interface Changes**
This does not modify functions prototypes, but add a bunch on structures that the developer can use directly.

**Additional Context**
I'll provide a pull request soon.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

treywelsh:
Some bug has been found by the ONE team in my code so I reopen this case and provide a pull request to fix this.

Add:
- Info method struct cleaning
- remove some mutualized part among structs, it cost a bit more of memory but the code is less tricky...

PR #2832, opened 2019-01-18
 F #2830: GOCA - migrate from xmlpath to encoding/xml package

treywelsh:
Add a bunch of structs, adapt Info methods with unmarshall, migrate tests and remove old pieces of xmlpath related code.
In addition, I add states to datastores and fix some typos.


treywelsh:
I just added the handling of the raft status.
Initially I used the zone.xsd which does not reflect the ouput of the command one.zone.info, but a mix of informations from one.zone.info with one.zone.raftstatus.
The documentation is not clear on this point.

Moreover the method one.zone.raftstatus is a command to execute on each front server to have the full raft status (If there is several fronts, with raft)

treywelsh:
I fixed some comments, and structure names. These are small changes.

treywelsh:
Some renaming, fixing some xml tags.

PR #2838, opened 2019-01-21
Support for snapshots without memory dumps on vcenter driver API

sergiojvg:


tinova:
Please use the F# xxx: <commit message> notation for commits

PR #2844, opened 2019-01-23
LXD fix

dann1:
- improved monitoring
- fix network driver for snaps


src/vmm_mad/remotes/lib/lxd/container.rb
```diff
@@ -162,16 +161,14 @@ def get_metadata
     # Runs command inside container
     # @param command [String] to execute through lxc exec
     def exec(command)
-        cmd = "#{LXC_COMMAND} exec #{@one.vm_name} -- #{command}"
+        cmd = "lxc exec #{@one.vm_name} -- #{command}"
```
rsmontero: I'd leave the constant for the command

Issue #2842, opened 2019-01-22
KVM probe of machines models stuck


vholer:
**Description**
Reading of the available machine/CPU models on KVM might stuck if the output from virsh is too long.

```
24702 ?        S      0:00 sh -c /var/tmp/one/im/kvm.d/../run_probes kvm-probes /var/lib/one//datastores 4124 60 0 centos7-ev-context-kvm-539f6-0.test 2>&1
24703 ?        S      0:00  \_ /bin/bash /var/tmp/one/im/kvm.d/../run_probes kvm-probes /var/lib/one//datastores 4124 60 0 centos7-ev-context-kvm-539f6-0.test
24710 ?        S      0:00      \_ /bin/bash /var/tmp/one/im/kvm.d/../run_probes kvm-probes /var/lib/one//datastores 4124 60 0 centos7-ev-context-kvm-539f6-0.test
24711 ?        S      0:00          \_ /bin/bash /var/tmp/one/im/kvm.d/../run_probes kvm-probes /var/lib/one//datastores 4124 60 0 centos7-ev-context-kvm-539f6-0.test
24746 ?        Sl     0:00              \_ ruby ./machines-models.rb kvm-probes /var/lib/one//datastores 4124 60 0 centos7-ev-context-kvm-539f6-0.test
24752 ?        S      0:00                  \_ virsh -r -c qemu:///system capabilities
```

Seeing those many many times:

```
$ ps -axf | egrep 'machines-models|virsh' | wc -l
63
```

Usage of `popen3` in the probe should be fixed to read the output in thread or changed to `capture3`:

https://github.com/OpenNebula/one/blob/5c59ff5894e375dbc977668d118810066e85ea51/src/im_mad/remotes/kvm-probes.d/machines-models.rb#L33
https://github.com/OpenNebula/one/blob/5c59ff5894e375dbc977668d118810066e85ea51/src/im_mad/remotes/kvm-probes.d/machines-models.rb#L98

**Details**
 - Affected Component: Drivers
 - Hypervisor: KVM
 - Version: 5.6, master

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

vholer:
PRs to consider:
- https://github.com/OpenNebula/one/pull/2848
[...]

PR #2848, opened 2019-01-25
B #2842: KVM probe of machines models stuck

vholer:


zakkg3:
why the libvirt url is hardcoded and not reading the sourced env var. 
this is not working then.
http://docs.opennebula.org/5.8/deployment/open_cloud_host_setup/kvm_driver.html#opennebula-configuration

lhttps://github.com/OpenNebula/one/blob/def9661b4aa7c708d51679eb051ce8bebce09d80/src/im_mad/remotes/kvm-probes.d/machines-models.rb#L33

Issue #2095, opened 2018-05-14
Force collectd to send metrics with HUP


OpenNebulaSupport:
# Enhancement Request
## Description

For example to update free disk space:

SIGHUP_COLLECTD="kill -HUP \`cat /tmp/one-collectd-client.p id\`" 
function collectd_monitor_force() { 
host=$1 
ssh_exec_and_log_no_error "$host" "$SIGHUP_COLLECTD" \ 
"Error forcing collectd update (Warning)" 
}

For example:

LVCREATE_CMD=$(cat <<EOF 
set -e 
mkdir -p $DST_DIR

rm -f "$DST_PATH" 
ln -sf "$DEV" "$DST_PATH" 
$SUDO $LVCREATE -L${SIZE}M -n $LV_NAME $VG_NAME 
EOF 
)

ssh_exec_and_log "$DST_HOST" "$LVCREATE_CMD" \ 
"Error creating $LV_NAME"

collectd_monitor_force "$DST_HOST"

for tm/ssh/clone

## Use case

Force updating metrics via the transfer manager driver

## Interface Changes

None

# Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

OpenNebulaSupport:
Implementation guideline

[collectd-client.zip](https://github.com/OpenNebula/one/files/2001082/collectd-client.zip)

xorel:
Also cli action should be created:
```
onehost forceupdate <host | id>
```

PR #2860, opened 2019-01-30
F #2095 Support persistent VM snapshots for kvm

xorel:


xorel:
tests: https://github.com/OpenNebula/development/pull/496

Issue #1699, opened 2018-01-30
Wild VM monitoring should not return datastores that contain only swap files


OpenNebulaSupport:
# Bug Report
## Version of OpenNebula

- [ ] Development build

## Component
- [ ] vCenter

## Description

If a vCenter VM features a swap file on a datastore, it is returned as the VCENTER_DS_REF (the first element of the datastore array is returned

### Expected Behavior

Datastore with non swap disk only should be returned

## How to reproduce

Create VM in vCenter with one disk in one datastore (ds1), and swap disk in another (ds2). Import only ds1 in OpenNebula. Try to import the Wild VM in OpenNebula

tinova:
Quick workaround (in virtual_machine.rb, vcenter vmm driver)
<         str << "VCENTER_DS_REF = \"#{@vm_info["datastore"].last._ref}\"\n"
---
>         str << "VCENTER_DS_REF = \"#{@vm_info["datastore"].first._ref}\"\n"

tinova:
Quick workaround (in virtual_machine.rb, vcenter vmm driver)

```
<         str << "VCENTER_DS_REF = \"#{@vm_info["datastore"].last._ref}\"\n"
---
>         str << "VCENTER_DS_REF = \"#{@vm_info["datastore"].first._ref}\"\n"
```

tinova:
Correct solution would be to add as datastore the one that holds the VM disk that boots the VM

tinova:
Added workaround ,postponing for 5.8

PR #2887, opened 2019-02-05
B #1699: Do not return datastores containing only SWAP

sergiojvg:


tinova:
- begin rescue
- function
- comments

src/vmm_mad/remotes/lib/vcenter_driver/vm_template.rb
```diff
@@ -895,6 +880,36 @@ def vm_to_one(vm_name)
...
+                        ds_ref = datastore._ref
+                        break
+                    end
+                end
+            else
+                if !@vm_info["datastore"]._ref.nil?
+                    ds_ref = @vm_info["datastore"]._ref
+                end
+            end
+        rescue Exception => e
```
tinova: It is a better practice to capture Standard Error

PR #2916, opened 2019-02-08
Fix "scp: error: unexpected filename: ."

quixoten:
OpenNebula was failing to enlist a new host. I made this change and restarted the nebula services and the host was then enlisted correctly.

```
Fri Feb  8 15:04:34 2019 [Z0][InM][I]: Command execution failed (exit code: 1): scp -rp /var/lib/one/remotes/. sd-prod-kvm1106:/var/tmp/one
Fri Feb  8 15:04:34 2019 [Z0][InM][I]: scp: error: unexpected filename: .
```

I was also able to reproduce this from the command line:

```
oneadmin@sd-prod-nebula1:~$ scp -rp /var/lib/one/remotes/. sd-prod-kvm1106:/var/tmp/one
scp: error: unexpected filename: .
```

### DIFF
--- a/src/im_mad/im_exec/one_im_exec.rb
+++ b/src/im_mad/im_exec/one_im_exec.rb
@@ -74,7 +74,7 @@ class InformationManagerDriver < OpenNebulaDriver
         if !action_is_local?(:MONITOR)
             if do_update == "1" || @options[:force_copy]
                 # Use SCP to sync:
-                sync_cmd = "scp -r #{@local_scripts_base_path}/. " \
+                sync_cmd = "scp -r #{@local_scripts_base_path}/* " \
                     "#{host}:#{@remote_scripts_base_path}"
 
                 # Use rsync to sync:
--- a/src/cli/one_helper/onehost_helper.rb
+++ b/src/cli/one_helper/onehost_helper.rb
@@ -363,7 +363,7 @@ class OneHostHelper < OpenNebulaHelper::OneHelper
                         sync_cmd = "rsync -Laz --delete #{REMOTES_LOCATION}" \
                                    " #{host['NAME']}:#{remote_dir}"
                     else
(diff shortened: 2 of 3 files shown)


rsmontero:
Notes:

Could be related to: https://nvd.nist.gov/vuln/detail/CVE-2018-20685#vulnCurrentDescriptionTitle

and this commit in OpenBSD scp.c

https://github.com/openbsd/src/commit/4667d15cf6214d4cb6b8c347f13864ca8c72002e

It seems that new versions of scp clients can be affected and hence monitoring drivers as described here.

Thanks for the heads up and patch!

src/im_mad/im_exec/one_im_exec.rb
```diff
@@ -74,7 +74,7 @@ def action_monitor(number, host, ds_location, do_update)
         if !action_is_local?(:MONITOR)
             if do_update == "1" || @options[:force_copy]
                 # Use SCP to sync:
-                sync_cmd = "scp -r #{@local_scripts_base_path}/. " \
+                sync_cmd = "scp -r #{@local_scripts_base_path}/* " \
```
vholer: If the destination directory `/var/tmp/one` is missing, the scp will fail (as the directory isn't explicitly created in this helper):

```
scp: /var/tmp/one: No such file or directory
```

There should be ssh+mkdir added.

```
sync_cmd = "ssh #{host['NAME']} mkdir -p '#{remote_dir}' 2>/dev/null &&" \
    "scp -rp #{REMOTES_LOCATION}/* " \
    "#{host['NAME']}:#{remote_dir} 2> /dev/null"
```

HeshamAboElMagd:
Adding /* to the scp or just remove /. , hadn't worked out for me.
Had to add the directory to be part of the scp as advised by @vholer .
Worked well with me . Thanks.

rsmontero:
We'll take care of mkdir dst after merge.....

PR #2917, opened 2019-02-11
F #1684

dann1:
- default image size set to 1GB
- create generic VM template for each app
- code rework
- minor improvements
- fix for *fstabless* images

### DIFF
--- a/src/market_mad/remotes/linuxcontainers/monitor
+++ b/src/market_mad/remotes/linuxcontainers/monitor
@@ -26,15 +26,23 @@ require 'rexml/document'
 #
 #-------------------------------------------------------------------------------
 class LinuxContainersMarket
+
     #---------------------------------------------------------------------------
     # Default Configuration parameters for the Driver
     #---------------------------------------------------------------------------
     DEFAULTS = {
-        :url    => 'https://images.linuxcontainers.org',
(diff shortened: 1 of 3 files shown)


src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -26,15 +26,23 @@ require 'rexml/document'
...
-        :sizemb => 2560,
-        :fs     => 'ext4',
+        :url => 'https://images.linuxcontainers.org',
+        :sizemb => 1024,
+        :fs => 'ext4',
         :format => 'raw',
-        :agent  => 'OpenNebula'
+        :agent => 'OpenNebula',
+        :template => {
+            :CPU => '1', :MEMORY => '768', :LXD_SECURITY_PRIVILEGED => 'true',
```
vholer: Just a Q:
1. why the templates are by default PRIVILEGED, the insecure ones?
2. shouldn't some of these LXD specific parameters be part of VM_RESTRICTED ?

dann1: yes it should, but the images are imported with the set of ids corresponding to the host root users, and currently, the way the drivers are, they don't perform id translation.

PR #2928, opened 2019-02-12
F #1684: Make LXD marketplace create VM templates

dann1:


src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -50,10 +58,9 @@ class LinuxContainersMarket
         @options.merge!(options)
 
         version_path = File.dirname(__FILE__) + '/../../VERSION'
+        value = "OpenNebula #{File.read(version_path)}"
 
-        if File.exists? version_path
-            @options[:agent] = "OpenNebula #{File.read(version_path)}"
-        end
+        @options[:agent] = value if File.exist? version_path
```
rsmontero: File.read needs to happen only if File.exists so this should be changed to:

@options[:agent] = "OpenNebula #{File.read(version_path)}" if File.exist? version_path

Otherwise the first File.read may throw an exception

src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -81,90 +88,89 @@ class LinuxContainersMarket
...
+                print_var(tmpl, 'SOURCE', source.to_s)
+                print_var(tmpl, 'IMPORT_ID', id)
+                print_var(tmpl, 'ORIGIN_ID', '-1')
+                print_var(tmpl, 'TYPE', 'IMAGE')
+                print_var(tmpl, 'PUBLISHER', 'linuxcontainers.org')
+                print_var(tmpl, 'FORMAT', 'raw')
+                print_var(tmpl, 'DESCRIPTION', description)
+                print_var(tmpl, 'VERSION', '1.0')
+                print_var(tmpl, 'TAGS', '')
+                print_var(tmpl, 'REGTIME', '1523362012')
```
rsmontero: Can we get a better time for REGTIME based on app metadata?

src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -190,6 +196,35 @@ class LinuxContainersMarket
 
         str << "#{name}=\"#{val}\"\n"
     end
+
+    def vm_tmpl64
```
rsmontero: This is static, move this to a class variable that initialize the base64 upon intialization

src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -190,6 +196,35 @@ class LinuxContainersMarket
...
+        template = JSON.generate(DEFAULTS[:template])
+        template = JSON.parse(template, :symbolize_names => true)
+        template = template_to_str(template)
+        Base64.strict_encode64(template)
+    end
+
+    def template_to_str(thash)
```
rsmontero: Remove. Keep the template in String form in the default section. No need to mange this two formats

src/market_mad/remotes/linuxcontainers/monitor
```diff
@@ -26,15 +26,23 @@ require 'rexml/document'
...
-        :sizemb => 2560,
-        :fs     => 'ext4',
+        :url => 'https://images.linuxcontainers.org',
+        :sizemb => 1024,
+        :fs => 'ext4',
         :format => 'raw',
-        :agent  => 'OpenNebula'
+        :agent => 'OpenNebula',
+        :template => {
+            :CPU => '1', :MEMORY => '768', :LXD_SECURITY_PRIVILEGED => 'true',
```
rsmontero: Change format see below

PR #2929, opened 2019-02-12
Feature 1684

dann1:


rsmontero:
We'll revisit some of this changes in the short term after 5.8, marketplace changes moved to other PR

PR #2938, opened 2019-02-14
Bug 2937

sergiojvg:


src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -105,7 +105,9 @@ for p in ${arguments//&/ }; do
     [ -n "$k" -a -n "$v" ] && eval $k=$v;
 done
 
-extension="tar.xz"
+#Only works with double extensions (.tar.xz and .tar.gz)
```
tinova: we need some error checking here, if the extension is not what we expect we should fail

Issue #2940, opened 2019-02-14
[GOCA] externalize client and modify the way to access to the resources (warning: breaking changes )


treywelsh:
**Description**
This pull request make some changes on the client, and the whole API.
/!\ It's an important breaking change /!\

As this imply breaking changes, I added some documentation (PR in the docs repository, and added some examples)

1. This pull request adds a controller notion.
It's a cheap layer added between the client and the entities API (host, vm ...).
It allows to rework the way to access to the resources via the API in a way that seems more intuitive to use.

For instance, to make requests, now, it's like this:
```
vm, err := ctrl.VM(id).Info()                              // Retrieve VM informations
err := ctrl.VM(id).Delete()                                // Delete a VM
err := ctrl.VMs().Create(template, boolean)    // Create a VM
vms, err := ctrl.VMs().Info()                            // Retrieve a list of VM informations (with less informations per instance)
id, err := ctrl.ByName("vm_name")            // Retrieve an ID by name
``` 

2.  The controller embed a client:
this pull request also externalize the client and allow to redefine it (For instance you may want to define a pool of clients to enabling some concurrency, or make a client that retry if the request fails ... ).
If the RPCCaller interface is implemented, the controller can use the external client.

3. By the way, I also added some other changes/fixes breaking the API:
    -  lock part to be more maintainable, add added some missing methods
    -  change parameters of some methods: permissions for chmod, typed parameter for Update methods
    -  ID are signed integers

**Use case**
Users of GOCA may want more flexibility on the client, and a more natural way to access the resources

**Interface Changes**
A great part of the API change.

**Additional Context**
I will provide a pull request in this way.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

eripa:
This is a very useful improvement. Any feedback from the OpenNebula team whether or not this is something that will be merged?

treywelsh:
Thank you very much for your feedback :)
It's not perfect, but it proposes some changes that we need.

I discuss with the OpenNebula member in charge of the Go binding, he seems busy but as far as I know it's being reviewed and some other pull request have been merged these days.

PR #2941, opened 2019-02-14
F #2940: GOCA - externalize the client and allow to extend it

treywelsh:
Firstly, I propose to externalize the client in providing a method NewClient, allowing the user to create new clients as they want and make request on the client they want.
To fully enable this, I add a pointer (in fact, an interface) to the client in each entity structure.

Secondly I define an interface for the GOCA client (a client has to define Call methods).
If a new defined client satisfy this interface, it can be used with all entities (VM, Host ...) to make the requests.

It imply important API changes.


treywelsh:
Please do not merge it for now, I work on an other idea, I may provide you an other pull request if I found it interesting.

Issue #2964, opened 2019-02-19
Wrong handling of timeout on provision create


vholer:
**Description**

Based on the error from testing:

```
11:14:00 2019-02-19 05:14:00 DEBUG : Command run: /var/lib/one/remotes/pm/packet/deploy /tmp/xmlDeploy120190219-31730-ezuhow TODO
11:15:47 2019-02-19 05:15:47 DEBUG : Command STDOUT: 4556ec5d-a717-40e2-8b58-3b5f0c8b7e37
11:15:47 2019-02-19 05:15:47 DEBUG : Command succeeded
11:44:08 2019-02-19 05:44:08 DEBUG : Command STDERR: ERROR MESSAGE --8<------
11:44:08 Deploy of Packet host with /tmp/xmlDeploy120190219-31730-ezuhow failed due to "Wait for host state 'active' timed out"
11:44:08 ["/usr/lib/one/ruby/packet_driver.rb:555:in `wait_state'", "/usr/lib/one/ruby/packet_driver.rb:123:in `deploy_host'", "/var/lib/one/remotes/pm/packet/deploy:41:in `<main>'"]"
11:44:08 
11:44:08 ERROR MESSAGE ------>8--
11:44:08 2019-02-19 05:44:08 WARN  : Command FAILED (code=255): /var/lib/one/remotes/pm/packet/deploy /tmp/xmlDeploy120190219-31730-ezuhow TODO
11:44:08 ERROR: Driver action '/var/lib/one/remotes/pm/packet/deploy' failed
11:44:08 Deploy of Packet host with /tmp/xmlDeploy120190219-31730-ezuhow failed due to "Wait for host state 'active' timed out"
11:44:08 /usr/lib/one/oneprovision/lib/driver.rb:44:in `>': comparison of Fixnum with nil failed (ArgumentError)
11:44:08 	from /usr/lib/one/oneprovision/lib/driver.rb:44:in `rescue in retry_loop'
11:44:08 	from /usr/lib/one/oneprovision/lib/driver.rb:37:in `retry_loop'
11:44:08 	from /usr/lib/one/oneprovision/lib/driver.rb:197:in `pm_driver_action'
11:44:08 	from /usr/lib/one/oneprovision/lib/provision.rb:349:in `block (2 levels) in deploy_hosts'
11:44:08 00:30:08     provision a new devices
11:44:08 00:00:00     list provisions
11:44:09 00:00:01     count clusters
11:44:09 00:00:00     count datastores
[...]

PR #2996, opened 2019-02-27
B #2964: Fix bug and add rescue to control exceptions

al3xhh:


al3xhh:
Applies to master and 5.8

PR #3030, opened 2019-03-05
b # 2993 delete langs and solved fix Vm_quota in conditions

jloboescalona2:


tinova:
Need to remove the files also from install.sh

Issue #3044, opened 2019-03-07
vcenter_downloader.sh fails to download vcenter images


sergiojvg:
**Description**
Wed Mar  6 20:57:16 2019 [Z0][MKP][E]: import: Command "/var/lib/one/remotes/market/http/../../datastore/downloader.sh -n vcenter://one/5/one-5.vmdk?param_dsid=104 /tmp/bba89b03ea8ed2a9e777abafadea11e3" failed: /usr/lib/ruby/vendor_ruby/builder/xchar.rb:111: warning: constant ::Fixnum is deprecated
Wed Mar  6 20:57:16 2019 [Z0][MKP][I]: Cannot download image /5/one-5.vmdk from datastore 104 Reason: "read image header failed"
Wed Mar  6 20:57:16 2019 [Z0][MKP][I]: ["/usr/lib/one/ruby/vcenter_driver/datastore.rb:551:in `is_descriptor?'", "/usr/lib/one/ruby/vcenter_driver/file_helper.rb:175:in `block in get_all_filenames_in_descriptor'", "/usr/lib/one/ruby/vcenter_driver/file_helper.rb:170:in `each'", "/usr/lib/one/ruby/vcenter_driver/file_helper.rb:170:in `get_all_filenames_in_descriptor'", "/usr/lib/one/ruby/vcenter_driver/file_helper.rb:200:in `dump_vmdk_tar_gz'", "/var/lib/one/remotes/datastore/vcenter_downloader.rb:51:in `<main>'"]

**To Reproduce**
Upload an image from a vCenter DS to an HTTP marketplace.

**Details**
 - Affected Component: [e.g. vCenter driver]

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

tinova:
Needs test

PR #3045, opened 2019-03-07
B #3044: Fix image downloader

sergiojvg:


tinova:
applied to master and one-5.8

Issue #1666, opened 2018-01-18
Marketplace driver expects always an image to import


vholer:
# Bug Report

## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [X] 5.4.5
- [X] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [x] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [X] Storage & Images
- [x] Sunstone
- [ ] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## Description

Even if the Marketplace appliance doesn't have any image specified (e.g. https://github.com/OpenNebula/marketplace/commit/af7429e152454f9c586e4efc29678f81f6dd1d22), the Marketplace driver always tries to import (non-existing) image and fails. To support the imageless appliances (e.g. for EC2 where the image is specified as `ami`), we must have an empty stub image specified (e.g. https://github.com/OpenNebula/marketplace/commit/978958b46cbf593c0378242d0b89f5d70369c5f8#diff-847b3ab2713a0327ce098c8e67454d37R39).

### Expected Behavior

Imageless appliances don't need to have an empty stub image.

### Actual Behavior

If there is no stub image specified, the import fails.

# Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

vholer:
This is more UI/CLI/core problem than the driver. Marketplace driver monitor can work even with imageless appliances, but the UI/CLI/core always do an invalid image import into the datastore.

![image](https://user-images.githubusercontent.com/414850/39190922-e9eec41a-47d5-11e8-8ff5-b5a31ecfdb6e.png)

PR #3065, opened 2019-03-12
B #1666: Marketplace driver expects always an image to import

al3xhh:
* Add new time of MarketPlaceApp image: NO_IMAGE
    * Modify the OCA, so if the type is NO_IMAGE no image is created
    * Modify Sunstone, so if the type is NO_IMAGE no ds is needed

### DIFF
--- a/include/MarketPlaceApp.h
+++ b/include/MarketPlaceApp.h
@@ -97,10 +97,10 @@ public:
      */
     enum Type
     {
-		UNKNOWN          = 0, /** < Unknown types     */
+        UNKNOWN          = 0, /** < Unknown types     */
         IMAGE            = 1, /** < Image MarketPlace App*/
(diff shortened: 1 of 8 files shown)


include/MarketPlaceApp.h
```diff
@@ -429,6 +445,12 @@ class MarketPlaceApp : public PoolObjectSQL
     {
         return new MarketPlaceAppTemplate;
     }
+protected:
+
+    /**
+     * MarketPlaceApp format with no images
+     */
+    const std::set<std::string> no_image = {"ami"};
```
vholer: I would propose not to depend on **ami** image format.
If this is the most straightforward way, let's use **external**, **remote**, or **cloud** keywords to identify no image to import.

al3xhh:
Applies to master and 5.8

PR #3073, opened 2019-03-13
Development: delete ARs before deleting the VNET, so the IPAM will be…

al3xhh:
… called


al3xhh:
Applies to 5.8 and master

Issue #3068, opened 2019-03-12
Provision create retries whole group of actions, not just the particular action


vholer:
**Description**

Having a simple provision template with cluster, 2 datastores, 3 vnets. If it fails when creating a vnet, retry action repeats again the create of already existing datastores and fails. Example output:

```
$ oneprovision create packet2.yaml -D
2019-03-12 15:42:20 INFO  : Creating provision objects
2019-03-12 15:42:20 DEBUG : Creating OpenNebula cluster: vh-f3043-devel
2019-03-12 15:42:20 DEBUG : cluster created with ID: 116
2019-03-12 15:42:20 DEBUG : Creating OpenNebula datastores: <%= @name %>-default
2019-03-12 15:42:20 DEBUG : Datastore created with ID: 334
2019-03-12 15:42:20 DEBUG : Creating OpenNebula datastores: <%= @name %>-system
2019-03-12 15:42:20 DEBUG : Datastore created with ID: 335
2019-03-12 15:42:20 DEBUG : Creating OpenNebula networks: nat
2019-03-12 15:42:20 DEBUG : Vnet created with ID: 33
2019-03-12 15:42:20 DEBUG : Creating OpenNebula vnets: private
2019-03-12 15:42:20 DEBUG : Vnet created with ID: 34
2019-03-12 15:42:20 DEBUG : Creating OpenNebula vnets: <%= @name %>-public
ERROR: Failed to create some resources
[one.vn.allocate] LQ==
1. quit
2. retry
3. skip
4. cleanup
Choose failover method:
2
2019-03-12 15:42:23 DEBUG : Creating OpenNebula datastores: vh-f3043-devel-default
ERROR: Failed to create some resources
[one.datastore.allocate] NAME is already taken by DATASTORE 334.
1. quit
2. retry
3. skip
4. cleanup
Choose failover method:
```

**Details**
 - Affected Component: Provision
 - Version: 5.8.1 beta

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

vholer:
BTW, in the output I can see also inconsistent logging:

- `2019-03-12 15:42:20 DEBUG : Creating OpenNebula datastores: <%= @name %>-default`
[...]

PR #3078, opened 2019-03-14
B #3068: Provision create retries whole group of actions, not just th…

al3xhh:
…e particular action

    * Change the retry loop to each resource
    * Change the message to show the evaluated erb


al3xhh:
Applies to 5.8 and master

PR #3100, opened 2019-03-20
Fix LXD storage errors on deploy start action error handling

### LINKED ISSUE #3098: LXD drivers fail to unmap non rootfs devices if deploy fails
**Description**
As seen in the [forum](https://forum.opennebula.org/t/problems-creating-lxd-containers-from-sunstone/7004/8) in [several entries](https://forum.opennebula.org/t/lxd-snap-awareness-for-ubuntu-version-18-10/7002/5) LXD drivers fail to deploy containers because the context disk was detected already mounted prior to the process of setting up the container storage. This condition occurs when, for any reason, the container fails to start and the error handling code fails to detect the device corresponding to the container context mountpoint.

**To Reproduce**
The idea is to create a container that fails to start and then trying to start it again.

For example, taking advantege of #3058 
- Create a container with an ovs nic
- poweroff the contaner
- start the container (this will fail and leave the context mounted)
- delete-recreate or retry (this will fail when detecting context already mounted)

**Expected behavior**
LXD drivers should revert all of the storage actions (maps and mounts) when a start fails for whatever reason

**Details**
 - Affected Component: Virtualization Drivers
 - Hypervisor: LXD
 - Version: 5.8.0

**Additional context**
When LXD attempts to start a container with declared disks in it devices section it changes the mountpoint of block devices, ex. a context device mapping with the container running is `/var/lib/lxd/devices/one-14/disk.context.context`

## Progress Status
- [ ] Branch created 
- [ ] Testing - QA
- [ ] Documentation 

dann1:
Fixes #3098

### DIFF
--- a/src/vmm_mad/remotes/lxd/deploy
+++ b/src/vmm_mad/remotes/lxd/deploy
@@ -51,7 +51,7 @@ if Container.exist?(container.name, client)
     container.get_metadata
 
     err_msg = 'A container with the same ID is already running'
-    raise LXDError, err_msg if container.status == 'Running'
+    raise err_msg if container.status == 'Running'
 
     container.config = config
 
--- a/src/vmm_mad/remotes/lxd/deploy
+++ b/src/vmm_mad/remotes/lxd/deploy
@@ -75,6 +75,8 @@ else
         operation = container.start
         raise operation if container.status != 'Running'
     rescue LXDError => exception
+        # TODO: Improve wait condition
+        sleep 1 # Wait for LXD to revert non-rootfs disks mountpoint
         storage_deleted = container.setup_storage('unmap')
 
         if storage_deleted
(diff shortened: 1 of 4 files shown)


src/vmm_mad/remotes/lxd/deploy
```diff
@@ -83,7 +85,7 @@ else
             OpenNebula.log_error 'failed to dismantle container storage'
         end
 
-        raise LXDError, exception
+        raise exception
```
rsmontero: Look for other `raise LXDError` occurrences

src/vmm_mad/remotes/lxd/deploy
```diff
@@ -74,7 +74,9 @@ else
     begin
         operation = container.start
         raise operation if container.status != 'Running'
-    rescue LXDError => exception
+    rescue StandardError => exception
+        # TODO: Improve wait condition
+        sleep 1 # Wait for LXD to revert non-rootfs disks mountpoint
```
rsmontero: As this is an exception we can be more aggressive in  the cool down of the FS and sleep for more seconds.

Issue #2808, opened 2019-01-15
Allow specifing blocksize for dd commands in tm drivers


kvaps:
**Description**

There is hardcoded value `bs=64k` for any dd command:

```
# grep -r 'bs=' /var/lib/one/remotes
/var/lib/one/remotes/tm/fs_lvm/mv:            "${DD} if=${SRC_DEV} of=${DST_DEV} bs=64k" \
/var/lib/one/remotes/tm/fs_lvm/resize:    ${DD} if=/dev/zero of="${DEV}" bs=64k \
/var/lib/one/remotes/tm/fs_lvm/clone:    #${DD} if=/dev/zero of="${DEV}" bs=64k \
/var/lib/one/remotes/tm/fs_lvm/ln:    ${DD} if=/dev/zero of="${DEV}" bs=64k \
/var/lib/one/remotes/tm/fs_lvm/cpds:    $DD if=$DEV of=$DST_PATH bs=64k
/var/lib/one/remotes/tm/fs_lvm/delete:        ${DD} if=/dev/zero of="\${DEV}" bs=64k || :
/var/lib/one/remotes/tm/fs_lvm/mkimage:    ${DD} if=/dev/zero of="${DEV}" bs=64k \
/var/lib/one/remotes/tm/fs_lvm/mvds:    $DD if=$DEV of=$DST_PATH bs=64k
/var/lib/one/remotes/datastore/fs/cp:    multiline_exec_and_log "set -e -o pipefail; $COPY_COMMAND | $SSH $DST_HOST $DD of=$TMP_DST bs=64k" \
/var/lib/one/remotes/datastore/ceph/cp:multiline_exec_and_log "set -e -o pipefail; $DUMP | $SSH $DST_HOST $DD of=$TMP_DST bs=64k" \
/var/lib/one/remotes/market/http/import:    multiline_exec_and_log "set -e -o pipefail; $CP_CMD | $SSH ${DST_HOST} $DD of=${DST_PATH} bs=64k" \
```

Need opportunity to override it with the some option

**Use case**
`bs=64k` - it's tiring long, I would like to override this value to `8M` or even to `16M` - it's much faster.

**Interface Changes**
Describe any changed you'd like to see to current interfaces including Sunstone, CLI or/and API  

**Additional Context**
Consider using some specific option like `DD_BS` or `DD_BLOCK_SIZE` for override default value,
I don't see big reason to specify it per datastore, so we can use some config files like:

```
[...]

PR #3116, opened 2019-03-24
F #2808: add DD_BLOCK_SIZE option

kvaps:
Add `DD_BLOCK_SIZE` option for allow specifying block size for the dd commands.

Fixes: https://github.com/OpenNebula/one/issues/2808

### DIFF
--- a/src/datastore_mad/remotes/fs/fs.conf
+++ b/src/datastore_mad/remotes/fs/fs.conf
@@ -14,15 +14,5 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
-# Default ZERO'ing behavior for the driver. When these variables are set the driver
-# will zero LVM space. This prevents data from a previous allocation to be exposed
-# in new volumes. However this will increase the VM allocation time.
-#
-#  Zero LVM volumes on creation or resizing
-ZERO_LVM_ON_CREATE=yes
-
-#  Zero LVM volumes on delete, when the VM disks are disposed
-ZERO_LVM_ON_DELETE=yes
-
 #  Block size for the dd commands
 DD_BLOCK_SIZE=64k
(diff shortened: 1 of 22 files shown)


src/datastore_mad/remotes/fs/fs.conf
```diff
@@ -0,0 +1,18 @@
+# -------------------------------------------------------------------------- #
```
rsmontero: Is this file included in install.sh?

src/market_mad/remotes/http/import
```diff
@@ -77,7 +77,7 @@ if [ -n "$BRIDGE_LIST" ]; then
     DST_HOST=`get_destination_host $ID`
     CP_CMD="$UTILS_PATH/downloader.sh -n '${IMPORT_SOURCE}' -"
 
-    multiline_exec_and_log "set -e -o pipefail; $CP_CMD | $SSH ${DST_HOST} $DD of=${DST_PATH} bs=64k" \
+    multiline_exec_and_log "set -e -o pipefail; $CP_CMD | $SSH ${DST_HOST} $DD of=${DST_PATH} bs=${DD_BLOCK_SIZE:-64k}" \
```
rsmontero: Where could this DD_BLOCK_SIZE set? Is there any conf file for market?

src/datastore_mad/remotes/fs/fs.conf
kvaps: nope

src/market_mad/remotes/http/import
kvaps: > Where could this DD_BLOCK_SIZE set
for now the only way to set is via global environment variables

> Is there any conf file for market?
Still not, but I can prepare it.

PR #3118, opened 2019-03-25
F #3099 Show container log when action fails

dann1:
the log is embedded into `wait?` method, so the rest of the containers methods benefit from it


dann1:
Applies to master as well

PR #3122, opened 2019-03-25
B #3119: Hourly sched actions just execute one time

al3xhh:
* Change next_action to return the time for next_action
    * Change XML query to use REPEAT
    * Update the TIME if the value is lower than DONE


al3xhh:
Applies to master and 5.8.1

Issue #3076, opened 2019-03-14
API - create one.vmpool.infoextended


futzi:
**Description**
In 5.8 the API call one.vmpool.info was changed to return only a subset of information of the XML. This is a good idea in general to speed up the API calls.
The issue is now: If you rely on additional information which is not returned you need additional API calls for each VM the additional information is required for.
This finally increases massively the time for external processes.

Perhaps it is possible to enhance the API in two ways (optionally)
- Create a one.vmpool.infoextended call which behaves as until 5.6 (return all information). This should be faster anyway if you create the right filters so you reduce the amount of data returned and also the overall number of API calls
- Alternatively allow adding the required data/xml as parameter

**Use case**
API driven additional interfaces/tools, monitoring etc.

**Interface Changes**
API: Add one.vmpool.infoextended call

**Additional Context**
none

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
I think the first option may be better. (At least straight-forward to implement)

I don't know the specific workflow of your application, but getting the whole list of VMs will sooner or later impact the response time. So I'd recommend either caching the pool (using the new proposed API call) or using the new search functionality in 5.8. You can list VMs with a specific attribute IP="10.0.0.1" for example, that would probably give you a much sorter list... 

[See the documentation here.
](http://docs.opennebula.org/5.8/operation/vm_management/vm_instances.html?highlight=search#searching-vm-instances)

futzi:
Hi Ruben - yes - I think, option 1 is the better way to go and also easy to be implemented on client side.
We are already caching the lists for a certain time  - and using the new filter it has boosted the calls really well - many thanx for this.

PR #3137, opened 2019-03-27
F #3076: create one.vmpool.infoextended

al3xhh:
* Add new function dump_extended
    * Add new API call one.vmpool.infoextended
    * Add parameter --extended in the CLI

### DIFF
--- a/src/rm/RequestManagerPoolInfoFilter.cc
+++ b/src/rm/RequestManagerPoolInfoFilter.cc
@@ -91,6 +91,28 @@ void VirtualMachinePoolInfo::request_execute(
         xmlrpc_c::paramList const& paramList,
         RequestAttributes& att)
 {
+
+    pool_filter(paramList, att, pool);
+}
+
+/* ------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------- */
+
+void VirtualMachinePoolInfoExtended::request_execute(
+        xmlrpc_c::paramList const& paramList,
+        RequestAttributes& att)
+{
+    pool_filter(paramList, att, pool);
+}
+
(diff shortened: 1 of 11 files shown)


src/rm/RequestManagerPoolInfoFilter.cc
```diff
@@ -615,6 +570,86 @@ void MarketPlacePoolInfo::request_execute(
...
         RequestAttributes& att)
 {
-    dump(att, ALL, -1, -1, "", "");
+    dump(att, ALL, -1, -1, "", "", false);
 }
 
+/* ------------------------------------------------------------------------- */
+/* ------------------------------------------------------------------------- */
+
+void RequestManagerPoolInfoFilter::pool_filter(
```
rsmontero: This Method should be for VirtualMachinePoolInfoFilter class

src/rm/RequestManagerPoolInfoFilter.cc
```diff
@@ -476,7 +418,8 @@ void RequestManagerPoolInfoFilter::dump(
         int                start_id,
         int                end_id,
         const string&      and_clause,
-        const string&      or_clause)
+        const string&      or_clause,
+        bool               extended)
```
rsmontero: Make extended a class attribute so we do not need to pass it around. At the end is a property of the class info method e.g.  VirtualMachinePoolInfoExtended vs  VirtualMachinePoolInfo. The base class can have a bool exteneded attribute set in the class constructor to configure its properties.

al3xhh:
Applies to master and 5.8

al3xhh:
Pushed directly to master

PR #3138, opened 2019-03-28
F #145: Dependencies for lxd market containers and bug fixes

sergiojvg:


src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -169,6 +169,10 @@ export PATH=\$PATH:/bin:/sbin
 
 echo "nameserver $DNS_SERVER" > /etc/resolv.conf
 
+yum install epel-release openssh-server -y > /dev/null 2>&1
+/usr/bin/echo "centos6" >> /root/version
```
vholer: What is the purpose of these `echo centos6` and 7?

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -169,6 +169,10 @@ export PATH=\$PATH:/bin:/sbin
...
+systemctl enable sshd
```
vholer: There is no systemd in C6 yet.

src/datastore_mad/remotes/lxd_downloader.sh
```diff
@@ -155,7 +155,7 @@ rm -f /etc/resolv.conf > /dev/null 2>&1
 echo "nameserver $DNS_SERVER" > /etc/resolv.conf
 
 apt-get update > /dev/null
-apt-get install curl -y > /dev/null 2>&1
+apt-get install curl openssh-server -y > /dev/null 2>&1
```
vholer: It's good to have DEBIAN_FRONTEND specified for the non-interactive runs. E.g.:
`DEBIAN_FRONTEND=noninteractive apt-get install curl openssh-server -y > /dev/null 2>&1`
￼

src/datastore_mad/remotes/lxd_downloader.sh
sergiojvg: ups, missed those debug lines

src/datastore_mad/remotes/lxd_downloader.sh
rsmontero: At the begning of the script there is a configuration section, please add there the extra packages needed:

```
#-------------------------------------------------------------------------------
# Downloader configuration attributes
#-------------------------------------------------------------------------------
...
PACKAGES_DEBIAN_UBUNTU="curl openssh-server"
PACKAGES_CENTOS6="epel-release openssh-server"
```

PR #3139, opened 2019-03-28
LXD Follow-up fixes

dann1:
- A simple check that helps with non opennebula managed shutdown operations (lxc stops, container inner poweroffs, sudden shutdowns)
- Sometimes we just need space


dann1:
Applies to master as well

Issue #3136, opened 2019-03-27
Timeout when using cleaunp option


al3xhh:
**Description**
When you have a provision with some hosts with running vms and the option cleanup is used, it can end in a timeout.

**To Reproduce**
Use cleanup option with several provisions and vms.

**Expected behavior**
The vms are deleted.

**Details**
 - Affected Component: Provision
 - Version: development and 5.8

**Additional context**
```
ERROR: Failed to delete running_vms
Timeout expired for state DONE. VM is in state ACTIVE, CLEANUP_DELETE
ERROR: Cleanup unsupported for this operation
ERROR: Failed to delete running_vms
Timeout expired for state DONE. VM is in state ACTIVE, CLEANUP_DELETE
ERROR: Cleanup unsupported for this operation
ERROR: Failed to delete running_vms
Timeout expired for state DONE. VM is in state ACTIVE, CLEANUP_DELETE
ERROR: Cleanup unsupported for this operation
ERROR: Failed to delete running_vms
Timeout expired for state DONE. VM is in state ACTIVE, CLEANUP_DELETE
ERROR: Cleanup unsupported for this operation
```

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #3142, opened 2019-03-28
B #3136: timeout when using cleaunp option

al3xhh:
* Increase the default timeout from 10 to 20
    * Add cleanup_timeout parameter to configure it via CLI
    * Implement wait_image_delete, to wait until images are deleted

### DIFF
--- a/src/oneprovision/lib/provision.rb
+++ b/src/oneprovision/lib/provision.rb
@@ -63,7 +63,8 @@ module OneProvision
         # Deletes the PROVISION
         #
         # @param cleanup [Boolean] True to delete running VMs and images
-        def delete(cleanup = false)
+        # @param timeout [Integer] Timeout for deleting running VMs
+        def delete(cleanup, timeout)
             Utils.fail('Provision not found.') unless exists
 
             if running_vms? && !cleanup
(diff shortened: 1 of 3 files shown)


al3xhh:
Applies to master and 5.8

src/oneprovision/lib/provision.rb
```diff
@@ -490,7 +496,30 @@ def delete_object(type, id)
...
+
+        # Waits until the image is deleted
+        #
+        # @param image   [OpenNebula::Image] Image to wait
+        # @param timeout [Integer ]          Timeout for delete
+        def wait_image_delete(image, timeout)
+            timeout.times do
+                rc = image.info
+
+                if OpenNebula.is_error?(rc) && rc.message =~ /Error getting/
```
vholer: I would say it's better to take any error as a successfull delete, instead of depending on some texts which might easily change in the future.

src/oneprovision/lib/provision.rb
```diff
@@ -63,7 +63,8 @@ def refresh
         # Deletes the PROVISION
         #
         # @param cleanup [Boolean] True to delete running VMs and images
-        def delete(cleanup = false)
+        # @param timeout [Integer] Timeout for deleting running VMs
+        def delete(cleanup = false, timeout = 20)
```
vholer: Does the `--cleanup-timeout` parameter really affect the `delete()` call?

PR #3156, opened 2019-04-01
Fix LXD market importer

dann1:
Replaces #3138 

- centos 6 and 7 enable sshd at startup after installing openssh-server
- alpine was missing curl, so it couldn't get the context package
- packages were being left on container `/root/` directory

Since we won't add ssh dependency for alpine on context-package, we could install openssh here on alpine case.


rsmontero:
Consider also my comment to #3148 about custom package installation. This commit does not include ssh installation?

dann1:
> Consider also my comment to #3138 about custom package installation. This commit does not include ssh installation?

Done, we had openssh stated but it was confirmed as a missing dep for context package. I guess until we release a new context version we can state here the missing deps such as openssh and dbus for ubuntu.

@vholer do you think stating openssh here on alpine it is correct, taking into account [my unmerged packages PR](https://github.com/OpenNebula/addon-context-linux/pull/149#issuecomment-477944124)

Issue #2981, opened 2019-02-21
Destroy VM isn't working with BRIDGE_LIST and UNDEPLOYED VMs


kvaps:
**Description**
When I removing undeployed VM, I have the next error:
```
[Z0][VM][I]: New state is ACTIVE
[Z0][VM][I]: New LCM state is EPILOG
[Z0][TM][I]: Command execution failed (exit code: 255): /var/lib/one/remotes/tm/fs_lvm/delete opennebula-oned-7789d84587-6n654:/var/lib/one//datastores/189/7273/disk.0 7273 100
```

Where `opennebula-oned-7789d84587-6n654` is my oned controller hostname.
But my system datastore `189` have `BRIDGE_LIST="node1 node2"` variable.

**To Reproduce**

I have two datastores with `fs_lvm`:

Images one:
```
NAME="Images Datastore"
ALLOW_ORPHANS="NO"
BRIDGE_LIST="node1 node2 node3 node4"
CLONE_TARGET="SYSTEM"
DRIVER="raw"
DS_MAD="fs"
DS_MIGRATE="YES"
LN_TARGET="SYSTEM"
RESTRICTED_DIRS="/"
SAFE_DIRS="/var/tmp"
TM_MAD="fs_lvm"
TYPE="IMAGE_DS"
```
System one:
```
NAME="System Datastore"
ALLOW_ORPHANS="NO"
BRIDGE_LIST="node1 node2"
DISK_TYPE="FILE"
DS_MIGRATE="YES"
RESTRICTED_DIRS="/"
SAFE_DIRS="/var/tmp"
SHARED="YES"
TM_MAD="fs_lvm"
TYPE="SYSTEM_DS"
```

Then I create VM from it. Then I do undeploy it hard, and then terminate this VM.

**Expected behavior**
VM successful removed.

**Details**
 - Affected Component: Core
 - Hypervisor: KVM
 - Version: 5.6.1

**Additional context**

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

kvaps:
@rsmontero please reopen the bug report the flowing patch does not solves problem fully. Sometimes opennebula can call `mv` script with whole directory target (like 5.8 do):

```
[...]

Issue #3130, opened 2019-03-26
LVM datastore migration does not work


kvaps:
**Description**

Here is specific check for that

https://github.com/OpenNebula/one/blob/release-5.8.0/src/tm_mad/fs_lvm/mv#L86-L88

```bash
if [ "${TYPE}" != "BLOCK" ]; then
    exit 0
fi
```

But disk type is always
```
TYPE = "FILE"
```

**To Reproduce**
* Configure lvm datastore according the [documentation](docs.opennebula.org/5.8/deployment/open_cloud_storage_setup/lvm_drivers.html)
* Instantiate new VM
* Shutdown it, and try to migrate datastore

**Expected behavior**
VM will be migrated with all data

**Actual behavior**
Only metadata directory will be moved, VM will continue use logical volume from the old datastore

**Details**
 - Affected Component: Storage drivers
 - Hypervisor: KVM
 - Version: 5.6 5.8

**Additional context**

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

kvaps:
There is more detailed discussion about this issue:
https://forum.opennebula.org/t/why-images-ds-parameters-affects-system-ds/7072

rsmontero:
Closing as duplicate of #2359

PR #3163, opened 2019-04-01
F #2352 #2359 #2981 #3130: Fix fs_lvm cleanup and offline migration issues

### LINKED ISSUE #2352: LVM cleanup executes on master instead storage node
**Description**
I just found out that `fs_lvm/delete` operation is executes on the main controller, but not in the compute node or any other node from BRIDGE_LIST 

**To Reproduce**
* Add `echo "$0" "$@" >> /tmp/123` line to the `fs_lvm/delete` file
* Create fs_lvm datastore with BRIDGE_LIST
* Create VM
* Remove VM
* Check `/tmp/123` on node and on container
* On controller you will see the next output:
```
/var/lib/one/remotes/tm/fs_lvm/delete opennebula:/var/lib/one//datastores/138/5826/disk.0 5826 100
```

**Expected behavior**

`opennebula` - it is my controller node.
I expected some other name instead, like name of compute node or name of node from BRIDGE_LIST.

```
/var/lib/one/remotes/tm/fs_lvm/delete <some_other_node>:/var/lib/one//datastores/138/5826/disk.0 5826 100
```

**Details**
 - Affected Component: Storage drivers
 - Hypervisor: KVM
 - Version: 5.6.0

**Additional context**

There is also cleanup procedure before creating and resizing drive, so, it is up to discussion is that really needed here to cleanup it twice.

## Progress Status
- [ ] Branch created 
- [ ] Testing - QA
- [ ] Documentation 

### LINKED ISSUE #2359: fs_lvm offline migration does not work
**Description**
No way for offline migrate VM with system drive on fs_lvm
VM is going to the FAILED state after migration, because lvm volume is not activated on target node.

Probably it is should be done by `tm/fs_lvm/premigrate` script, but it seems it is called only for online migration

**To Reproduce**
Try to offline migrate.

**Expected behavior**
VM succesful migrated

**Details**
 - Affected Component: Storage
 - Hypervisor: KVM
 - Version: 5.6.0

**Additional context**

For solve this problem need to activate volume after migration and before resuming VM:
```
lvchange -ay /dev/vg-one-137/lv-one-6446-0 
```

## Progress Status
- [ ] Branch created 
- [ ] Testing - QA
- [ ] Documentation 

kvaps:
#### Note for reviewer:

* **delete** action changes:

  *  Now it will take last VM's host, instead using host from BRIDGE_LIST 
     fixes #2352 *(but I still thinking that it is should be fixed on API level)*

  * Support for both path types, eg: `node:/var/lib/one//datastores/103/63/disk.0` and `node:/var/lib/one//datastores/103/63`. In second case it will zero and then remove all the lvm drives under VM directory recursively.
    fixes https://github.com/OpenNebula/one/issues/2981#issuecomment-478632787 and https://github.com/OpenNebula/one/issues/2981#issuecomment-478633894

* **mv** action changes:

  *  Remove check for `/VM/TEMPLATE/DISK[DISK_ID=$DISK_ID]/TYPE`. I don't see big reason in it. Perhaps it was attempt to define different mechanism for moving old volumes as files, if system datastore was converted from `shared` to `fs_lvm`. But I still thinking that it should also convert drive to use lvm instead file store during datastore migration. Besides, this check wasn't working anyway, this parameter always equal to `FILE`.
     fixes #2359 and #3130

  *  Added check for `/VM/LCM_STATE`. We don't need to execute `mv` on UNDEPLOY state.

Signed-off-by: kvaps <kvapss@gmail.com>

### DIFF
--- a/src/tm_mad/fs_lvm/delete
+++ b/src/tm_mad/fs_lvm/delete
@@ -41,67 +41,64 @@ source ${DRIVER_PATH}/../../datastore/libfs.sh
 # Return if deleting a disk, we will delete them when removing the
 # remote_system_ds directory for the VM (remotely)
 #-------------------------------------------------------------------------------
-DST_PATH=`arg_path $DST`
-DST_HOST=`arg_host $DST`
 
-DS_SYS_ID=$(echo $DST_PATH | grep -E '\/disk\.[[:digit:]]+$' | $AWK -F '/' '{print $(NF-2)}')
+# Workaround: take DST_HOST from the XML (#2352)
+while IFS= read -r -d '' element; do
+    XPATH_ELEMENTS[i++]="$element"
+done < <(onevm show -x $VM_ID | $XPATH \
+    /VM/HISTORY_RECORDS/HISTORY/HOSTNAME)
+unset i
+DST_HOST="${XPATH_ELEMENTS[i++]}"
+
+# Fallback take DST_HOST from the destination path
+DST_HOST="${DST_HOST:-$(arg_host $DST)}"
+DST_PATH=`arg_path $DST`
+
+if [ "${DST_PATH##*/}" = "$VM_ID" ]; then
+    DS_SYS_ID=$(echo $DST_PATH | $AWK -F '/' '{print $(NF-1)}')
+    FIND_DEPTH=1
+else
+    DS_SYS_ID=$(echo $DST_PATH | grep -E '\/disk\.[[:digit:]]+$' | $AWK -F '/' '{print $(NF-2)}')
+    FIND_DEPTH=0
+fi
 
 # Zero space
 ZERO_CMD=$(cat <<EOF
     set -x
-    DEV=\$(readlink $DST_PATH)
-
-    if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
-        ${DD} if=/dev/zero of="\${DEV}" bs=64k || :
-    fi
+    while read DISK_PATH; do
+        DEV=\$(readlink \$DISK_PATH)
+        if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
+            ${SUDO} ${LVCHANGE} -ay "\${DEV}"
+            ${DD} if=/dev/zero of="\${DEV}" bs=64k ||:
+        fi
+    done < <(find "$DST_PATH" -mindepth $FIND_DEPTH -maxdepth $FIND_DEPTH -type l)
 EOF
 )
 
 # Delete the device if it's a clone (LVM snapshot)
 DELETE_CMD=$(cat <<EOF
     set -x
-    DEV=\$(readlink $DST_PATH)
+    while read DISK_PATH; do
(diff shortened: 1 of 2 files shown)


kvaps:
I've added explanatory notes to the PR, please review.

kvaps:
Please do not merge yet.

There is need enhancements:
1) More strict check for datastore and volumes belongs to the VM
2) run `lvchange -an` during undeploy operation

kvaps:
OK, now ready for review

xorel:
Let's discuss the delete first:
- I don't think it's a good idea to get DST_HOST from the history (host might be down/unavailable already). Why we can't use the BRIDGE_LIST here?
- about the rest of the changes, I don't get the point. There are always 2 actions (if the VM has one disk). 1st deleting the disk (e.g. disk0) and second the directory. For the disk deletion, the DST_HOST is replaced by BRIDGE_LIST (if available) to prevent running on the frontend for the UNDEPLOYED host. For the directory removal, there should be no problem. For previously running VM this action will run on the hypervisor and for the UNDEPLOYED VM on the fronted (with previous `mv`).  What problem are the other changes addressing?

kvaps:
Hi,
> I don't think it's a good idea to get DST_HOST from the history (host might be down/unavailable already). Why we can't use the BRIDGE_LIST here?

There is few potential problems:

1) If datastore have no `BRIDGE_LIST`, the first `delete` action will always be called on the controller.
But if its failed, the second one (after the `recover --retry`) will always be called directly on the compute node, see https://github.com/OpenNebula/one/issues/2352#issuecomment-478741460
This fix unifies this behavior to always be run on the compute node.

2) Removing directory will never be called on the BRIDGE_LIST node, because this check will never return `DS_SYS_ID` for the `DST_PATH` which have no `disk\.[[:digit:]]+$` in the end.
https://github.com/OpenNebula/one/blob/3e1d571604d0ed3697a5e0145776b4e214b3abef/src/tm_mad/fs_lvm/delete#L46-L48

3) <s>OpenNebula 5.8 always call `delete` operation for the folder during vm termination, even if vm have the drives. https://github.com/OpenNebula/one/issues/2981#issuecomment-478632787.</s> My bad, I was wrong here!

4) We still have the node with activated LVM-device, it will continue existing there even after deletion.
Hoverer I solved it by adding this into `mv`:
https://github.com/OpenNebula/one/blob/0feb9d88cdbc0fa47e4fed265a18b290a2a70285/src/tm_mad/fs_lvm/mv#L90-L108

> What problem are the other changes addressing?

5) Remove check for `/VM/TEMPLATE/DISK[DISK_ID=$DISK_ID]/TYPE` solves problem with offline and datastore migration. fixes #2359 and #3163 

6) Added check for the `/VM/LCM_STATE` for deactivate lvm device during undeploy.

Any way. Since I was wrong about **3st point** and we have solution for **1st point**, I will prepare another PR for solve another problems only.

Thanks for review

kvaps:
Closing this PR, due new one: #3201

Issue #3080, opened 2019-03-14
Packet provision of CentOS 7 on c2.medium fails


al3xhh:
**Description**
When you try to deploy a provision in Packet with centos7 using c2.medium instance it fails to restart the network. So after restarting it, the machine losses the connection.

**To Reproduce**
Deploy a provision using this defaults:

```
defaults:
  connection: {}
  provision:
    driver:                 packet
    packet_token:    PACKET_TOKEN
    packet_project:  PROJECT_ID
    plan:                  c2.medium.x86
    os:                    centos7
```

**Expected behavior**
Deploy without errors.

**Details**
 - Affected Component: Provision
 - Version: development

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

vholer:
PRs to merge (it's safe to have it in 5.8.1):
- one:
  - PR https://github.com/OpenNebula/one/pull/3186 into **master** and cherry-pick into **one-5.8**
- docs:
  - PR https://github.com/OpenNebula/docs/pull/568 into **one-5.8-maintenance**
  - PR https://github.com/OpenNebula/docs/pull/569 into **master**

tinova:
Not merging the docs in master, since it won't be new in 5.10.

Closing.

PR #3186, opened 2019-04-04
B #3080: Packet provision of CentOS 7 on c2.medium fails

vholer:


tinova:
I guess this applies to one-5.8 as well?

PR #3201, opened 2019-04-08
 F #2352 #2359 #2981 #3130 #3233: Fix fs_lvm cleanup and offline migration issues

### LINKED ISSUE #2359: fs_lvm offline migration does not work

kvaps:
**Note for reviewer:**

This PR includes the next fixes:

1) Remove check for `/VM/TEMPLATE/DISK[DISK_ID=$DISK_ID]/TYPE` solves problem with offline and datastore migration. fixes #2359 and #3130. Add check for `LCM_STATE` instead, fs_lvm driver now correctly disables logical volume during undeploy VM.

2) Fix directory cleanup during termination undeployed VM for storage with the `BRIDGE_LIST` variable set. The expression for finding `DS_SYS_ID` was never work if `DST_PATH` is directory. fixes https://github.com/OpenNebula/one/issues/2981#issuecomment-478632787 and #2352 

5) Fix when node in `BRIDGE_LIST` have no activated lvm device, so zeroing is not working. Now it will always activate the device before zeroing.

### DIFF
--- a/src/tm_mad/fs_lvm/delete
+++ b/src/tm_mad/fs_lvm/delete
@@ -50,38 +50,49 @@ else
     DS_SYS_ID=$(echo $DST_PATH | grep -E '\/disk\.[[:digit:]]+$' | $AWK -F '/' '{print $(NF-2)}')
 fi
 
+# Activate device
+ACTIVATE_CMD=$(cat <<EOF
+    set -ex -o pipefail
+    if [ -L "$DST_PATH" ]; then
+        DEV=\$(readlink $DST_PATH)
+        if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
+            ${SUDO} ${SYNC}
+            ${SUDO} ${LVSCAN}
+            ${SUDO} ${LVCHANGE} -ay "\${DEV}"
+        fi
+    fi
+EOF
+)
+
 # Zero space
 ZERO_CMD=$(cat <<EOF
     set -ex -o pipefail
-    DEV=\$(readlink $DST_PATH)
-
-    if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
-        ${SUDO} ${SYNC}
-        ${SUDO} ${LVSCAN}
-        ${SUDO} ${LVCHANGE} -ay "\${DEV}"
-        ${DD} if=/dev/zero of="\${DEV}" bs=64k || :
+    if [ -L "$DST_PATH" ]; then
+        DEV=\$(readlink $DST_PATH)
+        if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
+            ${DD} if=/dev/zero of="\${DEV}" bs=64k || :
+        fi
     fi
 EOF
 )
 
 # Delete the device if it's a clone (LVM snapshot)
 DELETE_CMD=$(cat <<EOF
-    set -x
-    DEV=\$(readlink $DST_PATH)
+    set -ex -o pipefail
 
     if [ -d "$DST_PATH" ]; then
         rm -rf "$DST_PATH"
-    else
-        rm -f $DST_PATH
+        exit 0
+    fi
 
-        if [ -z "\$DEV" ]; then
-            exit 0
-        fi
-
+    if [ -L "$DST_PATH" ]; then
+        DEV=\$(readlink $DST_PATH)
         if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
-            $SUDO $LVREMOVE -f \$DEV
+            $SUDO $LVREMOVE -f "\$DEV"
         fi
     fi
+
+    rm -f "$DST_PATH"
 EOF
 )
 
@@ -101,6 +112,10 @@ if [ -n "${DS_SYS_ID}" ]; then
     fi
 
     if [ "${ZERO_LVM_ON_DELETE}" = "yes" ]; then
+        LOCK="tm-fs_lvm-${DS_SYS_ID}.lock"
+        exclusive "${LOCK}" 120 ssh_exec_and_log "$DST_HOST" "$ACTIVATE_CMD" \
+            "Error activating disk $SRC_PATH"
+
         ssh_exec_and_log "$DST_HOST" "$ZERO_CMD" "Error cleaning $DST_PATH"
     fi
 
--- a/src/tm_mad/fs_lvm/delete
+++ b/src/tm_mad/fs_lvm/delete
@@ -52,10 +52,13 @@ fi
 
 # Zero space
 ZERO_CMD=$(cat <<EOF
-    set -x
+    set -ex -o pipefail
     DEV=\$(readlink $DST_PATH)
 
     if echo "\$DEV" | grep "^/dev/" &>/dev/null; then
+        ${SUDO} ${SYNC}
+        ${SUDO} ${LVSCAN}
+        ${SUDO} ${LVCHANGE} -ay "\${DEV}"
         ${DD} if=/dev/zero of="\${DEV}" bs=64k || :
     fi
 EOF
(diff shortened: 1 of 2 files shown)


kvaps:
@xorel please review the changes

src/tm_mad/fs_lvm/mv
```diff
@@ -77,22 +78,37 @@ if [ `is_disk $DST_PATH` -eq 1 ]; then
...
         XPATH_ELEMENTS[i++]="$element"
     done < <(onevm show -x $VMID | $XPATH  \
-                        /VM/TEMPLATE/DISK[DISK_ID=$DISK_ID]/TYPE )
+                        /VM/LCM_STATE )
 
-    TYPE="${XPATH_ELEMENTS[j++]}"
-
-    ####
-
-    if [ "${TYPE}" != "BLOCK" ]; then
```
xorel: I don't understand this yet. It still looks to me like can-not-happen case, once it passed the condition on line 67 (is_disk) it should always be "BLOCK", right?

src/tm_mad/fs_lvm/delete
```diff
@@ -44,37 +44,55 @@ source ${DRIVER_PATH}/../../datastore/libfs.sh
 DST_PATH=`arg_path $DST`
 DST_HOST=`arg_host $DST`
 
-DS_SYS_ID=$(echo $DST_PATH | grep -E '\/disk\.[[:digit:]]+$' | $AWK -F '/' '{print $(NF-2)}')
+if [ "${DST_PATH##*/}" = "$VM_ID" ]; then
+    DS_SYS_ID=$(echo $DST_PATH | $AWK -F '/' '{print $(NF-1)}')
+else
+    DS_SYS_ID=$(echo $DST_PATH | grep -E '\/disk\.[[:digit:]]+$' | $AWK -F '/' '{print $(NF-2)}')
```
xorel: I think this else clause is not needed, using empty "DS_SYS_ID" later (line 99) it detects if it's going to delete LV or just directory. Then also the changes in the DELETE_CMD are not required.

kvaps: We need "DS_SYS_ID" in both cases:
1. If we removing disk, we should go on the BRIDGE_LIST node and remove the disk.
2. If we removing directory, we should go on the BRIDGE_LIST node and remove the directory.

If we would not do that, it will try to remove directory on the controller, what cause an error if there is no mounted datastore

xorel: OK, clear, then could you also remove the `if` statement on line 99 + it's else 125..127?

src/tm_mad/fs_lvm/mv
kvaps: I presume this check was added there for the case if datastore was converted from `shared` to `fs_lvm`, as described in [documentation](http://docs.opennebula.org/5.8/deployment/open_cloud_storage_setup/lvm_drivers.html):
> It is a good idea to first deploy a shared FileSystem datastore and once it is working replace the associated System datastore with the LVM one maintaining the shared mount point as described below.

There two things about that:

1) Check is not working, because `/VM/TEMPLATE/DISK[DISK_ID=$DISK_ID]/TYPE` always equals `FILE`, I presume it might be working if we would replace it to `DISK_TYPE` which is inherit from the datastore, but in this case we lose opportunity to change it on the fly. Once this parameter set for the image it can't be updated in the future.
2) Better to check it on the place, because it also adds opportunity for convert simple file to the block device by simple migrating datastore via sunstone.
https://github.com/OpenNebula/one/blob/5a4e6633223fdd67c5ac0205c41339b559fb864c/src/tm_mad/fs_lvm/delete#L57-L58

src/tm_mad/fs_lvm/delete
kvaps: ok, done

kvaps:
> Could you please also rebase to get rid of the conflict with the missing DD_BLOCK_SIZE?

Done, I've also replaced DST_PATH check with internal `is_disk` function:

https://github.com/OpenNebula/one/blob/515ee094305cefb4b5a2f5007fb905f69a45e79a/src/tm_mad/fs_lvm/delete#L47-L51

kvaps:
Found the problem, when we executing `delete` operation on the `BRIDGE_LIST` node, device if not removing from the device mapper on the current node, where volume was activated

xorel:
That's good point.

OK, I did some tests and realized that detecting the UNDEPLOYED using the LCM check `10|30|41|42` is wrong, when the `fs_lvm/delete` is called the `LCM_STATE` is always 11 (EPILOG) no matter the previous state (except some error states), that was introduced by me by mistake. 

So, I am afraid the only way how to detect if the `BRIDGE_LIST` should be used is tho compare 
 the `$DST_HOST` and the `HOST` from the `VM` history, like you did before.

And third point, it seems that we are completely missing the device mapper clanup, I think we shoul call `dmsetup remove` after each `lvremove`.

This thing now bloated quite a lot and I am thinking starting again from scratch and tackle the issues one by one.

kvaps:
> using the LCM check 10|30|41|42 is wrong, when the fs_lvm/delete is called the LCM_STATE is always 11 (EPILOG) no matter the previous state (except some error states), that was introduced by me by mistake.

The check is working fine, it was not working because you've tested it with wrong action. You've tested `fs_lvm/delete` but check was added into `fs_lvm/mv`action :)

By the way, it is working really fine, I've already applied this path on production.

> And third point, it seems that we are completely missing the device mapper clanup, I think we shoul call dmsetup remove after each lvremove.

We don't have to, because it is automatically done by `lvremove` command.
This is just should be run on the same host where VM is located, https://github.com/OpenNebula/one/commit/35dfb51bdb3fabfda49cb21131223f55d72bf61d does this.

kvaps:
> using the LCM check 10|30|41|42 is wrong, when the fs_lvm/delete is called the LCM_STATE is always 11 (EPILOG) no matter the previous state (except some error states), that was introduced by me by mistake.

You're right, this check is working fine for the `fs_lvm/mv` action, but it is wrong to use it for delete.
Let me fix that.

> And third point, it seems that we are completely missing the device mapper clanup, I think we shoul call dmsetup remove after each lvremove.

We don't have to, because it is automatically done by `lvremove` command.
This is just should be run on the same host where VM is located, https://github.com/OpenNebula/one/commit/35dfb51bdb3fabfda49cb21131223f55d72bf61d does this.

kvaps:
@xorel found elegant solution: `[ "$LAST_HOST" != "$DST_HOST" ]` - means running on frontend.

Commit (https://github.com/OpenNebula/one/pull/3201/commits/e429d1cf4f12b274fface946d247d46b310626b9) force pushed to the branch for replace last wrong changes.

xorel:
> @xorel, I've found elegant solution: `[ "$LAST_HOST" != "$DST_HOST" ]` - means running on frontend.
> 

OK, this seems to work better.

One more detail, when testing I noticed that after offline migration the LV remains active on the SRC_HOST, I suppose the `mv` action should always deactivate the LV and only if the action is `stop` or `undeploy` 
 then exit. I don't see any other scenarios when the SRC LV should not be deactivated.

kvaps:
Indeed

kvaps:
Found another problem: **undeploy** --> deploy on selected datastore is not working

kvaps:
> > after offline migration the LV remains active on the SRC_HOST, I suppose the `mv` action should always deactivate the LV and only if the action is `stop` or `undeploy`
> > then exit.
> 
> Indeed
> 
> > I don't see any other scenarios when the SRC LV should not be deactivated.
> 
> When copying from the frontend (deploy)

solved

Issue #2505, opened 2018-10-15
Nested filters not working.


sergiojvg:
**Description**
When several -f filtering options are specified, a logical OR is used for matching. It should be switched to an AND.

**To Reproduce**
onehost list -f RVM=41,CLUSTER=default
onehost list -f CLUSTER=default -f RVM=23

**Expected behavior**
RVM=41 AND CLUSTER=default

**Implementation**
A new option will be created to use AND instead of OR.

**Details**
 - Hypervisor: KVM
 - Version: 5.4, 5.6

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

sergiojvg:
https://github.com/OpenNebula/docs/pull/563 --> into one/5.8-maintenance
https://github.com/OpenNebula/one/pull/3083 --> into master
https://github.com/OpenNebula/development/pull/571 --> into master

PR #3240, opened 2019-04-16
F #2505: Fix detach disk with snapshots for vcenter

sergiojvg:


tinova:
adding to master and one-5.8

Issue #2451, opened 2018-09-25
Make pre/post/clean steps easily extensible


rsmontero:
**Description**

Network configuration is performed in two main steps: before creating the VM (pre) and after it (post). This issue is also to make easily extensible those steps.

Drivers will have an optional pre.d and post.d directories. Any additional file in these directories will be executed after the main action is performed, following (or using) the interface of run-parts command.

**Use case**
Use case is to easily modify or extend current drivers; and so easily integrate them with other network components or SDN's.

**Interface Changes**
New directories for custom actions. Changes should be compatible with current layout (i.e. missing pre.d or post.d dirs)

**Additional Context**
None

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

dann1:
PRs:

- https://github.com/OpenNebula/infra/pull/229
- https://github.com/OpenNebula/development/pull/594
- https://github.com/OpenNebula/one/pull/3199

PR #3255, opened 2019-04-22
F #2451 Extend dummy vnm

dann1:


src/vnm_mad/remotes/dummy/clean
```diff
@@ -16,4 +16,8 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
-exit 0
+stdin="$(</dev/stdin)"
+
+for f in "$0".d/* ; do
+  echo "$stdin" | ./"$f" "$@"  || exit 1
```
rsmontero: Probably missing
```
[ -x $f ] && echo ....
```

dann1: Done, I added `|| exit 0`, else it would run exit 1

PR #3260, opened 2019-04-23
LXD Minor improvements

dann1:


src/vmm_mad/remotes/lib/lxd/client.rb
```diff
@@ -130,7 +130,14 @@ def get_response(request, data)
 
         response.reading_body(@socket, request.response_body_permitted?) {}
 
-        response = JSON.parse(response.body)
+        begin
+            response = JSON.parse(response.body)
+        rescue StandardError => e
```
rsmontero: Catch the exception you are interested in instead of match on error messages

begin
...
rescue JSON::ParserError
...
end

src/vmm_mad/remotes/lib/lxd/client.rb
```diff
@@ -130,7 +130,14 @@ def get_response(request, data)
...
+            raise e unless e.message.include?('A JSON text must at least'\
```
rsmontero: Would it better to raise a LXDError as below? Now we are just triggering the JSON parse error

src/vmm_mad/remotes/lib/lxd/client.rb
```diff
@@ -130,7 +130,14 @@ def get_response(request, data)
...
+                'contain two octets! (JSON::ParserError)')
+
+            response.reading_body(@socket, request.response_body_permitted?) {}
```
rsmontero: If it is other error we are reading again, but the resulting `response` is not parsed

src/vmm_mad/remotes/lib/lxd/client.rb
```diff
@@ -119,19 +119,24 @@ def socket(lxd_path)
...
 
-        request.exec(@socket, '1.1', request.path)
-
         response = nil
 
-        loop do
-            response = Net::HTTPResponse.read_new(@socket)
-            break unless response.is_a?(Net::HTTPContinue)
-        end
+        5.times do
```
rsmontero: * Add a constant in the LXD in the LXDCLient class (API_RETRY) and change loop for `API_RETRY.times`
 * A break is missing in case of success to not do all the retries if it is not necessary
* if possible raise LXDError, better with a message ( `raise LXDError('Failed to read response from LXD'`)

rsmontero: * Use a class constant API_RETRY, then API_RETRY.times
* A break statement is missing to not perform uneeded retries
* If possible raise LXDError(`Failed to read response from LXD`)

PR #3270, opened 2019-04-25
F #3272: extend vmm/save and vmm/restore

atodorov-storpool:
so both will have pre and post hooks as follow

save.<tm_mad>-pre
resotre.<tm_mad>-post

Signed-off-by: Anton Todorov <atodorov-storpool@users.noreply.github.com>


rsmontero:
Added a related feature #3272

atodorov-storpool:
I am closing #3270 and #3271 in favour of #3272 because this solution could be implemented faster and at the moment with a configuration change that does not violate the OpenNebula Polices.

Issue #3272, opened 2019-04-25
Add presave/postsave and prerestore/postrestore tm actions


kvaps:
**Description**
We need this functional for save and restore checkpoint file directly into storage, without saving it into some place before

**Use case**
Faster save and restore and migrate operations for storage drivers like ceph. And no requirement for additional space to save temprorary file

**Interface Changes**
* All TM_drivers should implement 4 new actions: `presave`, `postsave`, `prerestore`, `postrestore`, most of them can be just do `exit 0`
* Remove `save.<TM_MAD>` and `resotre.<TM_MAD>` hooks from vmm-driver

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

atodorov-storpool:
I am closing #3270 and #3271 in favour of #3272 because this solution could be implemented faster and at the moment with a configuration change that does not violate the OpenNebula Polices.

kvaps:
This is still desirable change.

atodorov-storpool:
Second for this.
It is already integrated into addon-storpool and is working flawlessly in production setups.

Issue #3271, opened 2019-04-25
extend save/restore on KVM VMM_MAD to have both pre and post hooks


atodorov-storpool:
**Description**
Brief description of the new functionality
Extend VMM_MAD save/restore to call tm_mad pre and post scripts

**Use case**
On System datastores with TM_MAD backed by block devices it is possible to store the checkpoint file directly on block device, skipping the step of storing the file on Host filesystem first.

**Interface Changes**
Extend save and restore on VMM_MAD to have both "pre" and "post" hooks to SYSTEM DS's TM_MAD. With the change the _vmm/save_ script will look for _vmm/save.<tm_mad>-pre_ and _vmm/restore_ will look for _vmm/restore.<tm_mad>-post_. Both pieces of code are borrowed from the existing parts of current save/restore

**Additional Context**
Following the [forum discussion](https://forum.opennebula.org/t/shared-mode-for-the-ceph-based-system-datastires/7164/5)

I think that the ceph TM_MAD could be updated to use this functionality but have no ceph cluster to test it, though.

Pull request #3270

There is alternative solution by @kvaps introduced in #3272 

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

kvaps:
link https://github.com/OpenNebula/one/issues/3272

atodorov-storpool:
I am closing #3270 and #3271 in favour of #3272 because this solution could be implemented faster and at the moment with a configuration change that does not violate the OpenNebula Polices.

PR #3273, opened 2019-04-25
F #3272: Add presave postsave prerestore postrestore tm actions [not completed] [need help]

kvaps:
Signed-off-by: kvaps <kvapss@gmail.com>

### DIFF
--- a/include/TransferManager.h
+++ b/include/TransferManager.h
@@ -210,6 +210,26 @@ public:
         ostream&                xfr);
 
     /**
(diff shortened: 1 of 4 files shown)


rsmontero:
Related to #3270. We'll address both at the same time

atodorov-storpool:
I am closing #3270 and #3271 in favour of #3272 because this solution could be implemented faster and at the moment with a configuration change that does not violate the OpenNebula Polices.

PR #3293, opened 2019-04-29
F #3293 Read driver action on attach_disk using STDIN for vcenter dri…

sergiojvg:
…vers


sergiojvg:
--> master
--> one-5.8

PR #3304, opened 2019-05-06
OpenvSwitch “del_flows” override loop variables

baby-gnu:
The “del_flows” override the loop variable “m” when matching the port,
on the second loop which result in the following error.

    /var/tmp/one/vnm/ovswitch/OpenvSwitch.rb:287:in `match': wrong argument type MatchData (expected Regexp) (TypeError)

* src/vnm_mad/remotes/ovswitch/OpenvSwitch.rb (OpenvSwitchVLAN#del_flows):
  use two different variables


tinova:
don't mind the GitCop on this, no idea why it is complaining

Issue #2453, opened 2018-09-27
Catch monitor errors


joseangelgm:
**Description**

#<SystemExit: exit>
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:34:in `exit'
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:34:in `return_if_error'
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:74:in `one_item'
/usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:3327:in `get_vm'
/usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:3390:in `new_from_ref'
/usr/lib/one/ruby/vcenter_driver/host.rb:482:in `block in monitor_vms'
/usr/lib/one/ruby/vcenter_driver/host.rb:459:in `each'
/usr/lib/one/ruby/vcenter_driver/host.rb:459:in `monitor_vms'
./poll:56:in `<main>'
[one.vm.info] Error getting virtual machine [205].
#<SystemExit: exit>
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:34:in `exit'
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:34:in `return_if_error'
/usr/lib/one/ruby/vcenter_driver/vi_helper.rb:74:in `one_item'
/usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:3327:in `get_vm'
/usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:3390:in `new_from_ref'
/usr/lib/one/ruby/vcenter_driver/host.rb:482:in `block in monitor_vms'
/usr/lib/one/ruby/vcenter_driver/host.rb:459:in `each'
/usr/lib/one/ruby/vcenter_driver/host.rb:459:in `monitor_vms'
./poll:56:in `<main>'

**To Reproduce**
Execute manually vcenter poll

**Expected behavior**
Catch this error to not display them

**Details**
 - Affected Component: tm mad
 - Hypervisor: vCenter
 - Version: 5.6.2

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

sergiojvg:
[...]

PR #3319, opened 2019-05-08
F #2453:  Catch monitor errors

sergiojvg:


sergiojvg:
--> master

src/vmm_mad/remotes/lib/vcenter_driver/host.rb
```diff
@@ -344,199 +344,203 @@ def monitor_host_systems
...
-            str_info << vm_info
-        end
+            view.DestroyView # Destroy the view
 
-        view.DestroyView # Destroy the view
-
-        return str_info, last_mon_time
+            return str_info, last_mon_time
+        
+        rescue
```
tinova: Needs to rescue only StandardError

src/vmm_mad/remotes/lib/vcenter_driver/host.rb
```diff
@@ -344,199 +344,203 @@ def monitor_host_systems
...
+            return "", ""
```
tinova: Shouldn't it add error_monitoring as well?

tinova:
not needed, already solved #2453

Issue #3326, opened 2019-05-10
Imported templates can be forced to use a specific MAC.


sergiojvg:
Description
Imported templates using an existing VNET will be locked to use a specific MAC address.

Details
    Affected Component: vCenter driver
    Hypervisor: vCenter
    Version: development

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

sergiojvg:
PR: https://github.com/OpenNebula/one/pull/3327 --> master, --> one-5.8

PR #3327, opened 2019-05-10
B #3326: Do not set mac when importing template

sergiojvg:


sergiojvg:
--> master
--> one-5.8

Issue #3331, opened 2019-05-13
Sunstone support token configuration not working behind a proxy


OpenNebulaSupport:
**Description**
When Sunstone is configured to use a proxy server, support token configuration is not using that proxy. 

**Details**
 - Affected Component: Sunstone
 - Version: 5.8.1

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

sergiojvg:
https://github.com/OpenNebula/one/pull/3332 --> master, --> one-5.8

PR #3332, opened 2019-05-14
B #3331: Support token behind proxy

sergiojvg:


sergiojvg:
--> master
--> one-5.8??

Issue #3337, opened 2019-05-14
Implement retry on vCenter shutdown operation


sergiojvg:
**Description**
Some vCenter drivers (like shutdown) can fail with the following error:

```
Mon May 13 21:10:51 2019 [Z0][VM][I]: New LCM state is SHUTDOWN
Mon May 13 21:10:58 2019 [Z0][VMM][I]: Command execution failed (exit code: 255): /var/lib/one/remotes/vmm/vcenter/shutdown 'vm-*' '*' 16 *
Mon May 13 21:10:58 2019 [Z0][VMM][E]: shutdown: Shutdown of VM vm-* on vCenter cluster Cluster failed due to "SystemError: A general system error occurred: Invalid fault"
Mon May 13 21:10:58 2019 [Z0][VMM][I]: /usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:2539:in `rescue in shutdown'
Mon May 13 21:10:58 2019 [Z0][VMM][I]: /usr/lib/one/ruby/vcenter_driver/virtual_machine.rb:2531:in `shutdown'
Mon May 13 21:10:58 2019 [Z0][VMM][I]: /var/lib/one/remotes/vmm/vcenter/shutdown:64:in `<main>'
Mon May 13 21:10:58 2019 [Z0][VMM][I]: Failed to execute virtualization driver operation: shutdown.
```

This should be fixed using a retry.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

sergiojvg:
PR: https://github.com/OpenNebula/one/pull/3338 --> master, --> one-5.8

PR #3338, opened 2019-05-14
F #3337: Retry if failure during shutdown

sergiojvg:


sergiojvg:
--> master
--> one-5.8

Issue #3362, opened 2019-05-22
onedb fsck remove nic alises from AR leases


atodorov-storpool:
**Description**
```bash
[oneadmin@vs04 ~]$ onedb fsck -s one.db 
Sqlite database backup stored in /var/lib/one/one.db_2019-5-22_13:41:59.bck
Use 'onedb restore' to restore the DB.

VNet 1 AR 0 has leased 10.2.85.11 to VM 16, but it is actually free
VNet 1 AR 0 has leased 10.2.85.12 to VM 13, but it is actually free
VNet 1 AR 0 has leased 10.2.85.13 to VM 13, but it is actually free
VNet 1 AR 0 has leased 10.2.85.14 to VM 13, but it is actually free
VNet 1 AR 0 has leased 10.2.85.15 to VM 16, but it is actually free
VNet 1 AR 0 has leased 10.2.85.17 to VM 15, but it is actually free
VNet 1 has 8 used leases, but it is actually 3

Total errors found: 7
Total errors repaired: 7
Total errors unrepaired: 0
A copy of this output was stored in /var/log/one/onedb-fsck.log
```

**To Reproduce**
* Add IP aliases to a VM nic
* Stop opennebula service
* Run `onedb fsck`

The leases of NIC's alias IPs are removed (see the snippet in the description)

**Expected behavior**
The NIC's alias IPs should not be removed by fsck

**Details**
 - Affected Component: [onedb]
 - Hypervisor: [KVM]
 - Version: [5.8.2]

**Additional context**
A pull request is on the way...

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #3363, opened 2019-05-22
B #3362: count NIC aliases in onedb fsck

atodorov-storpool:


atodorov-storpool:
signed-off conflicts with gitcop
Was using `Signed-off-by: Anton Todorov <atodorov-storpool@users.noreply.github.com>` but gitcop doesn't like it :/ 
will try with a.todorov@storpool.com

Issue #3373, opened 2019-05-27
live migrate failing due to "Arguments list too long"


atodorov-storpool:
**Description**
When the six-th argumet to {pre,post,fail}migrate is too log (it is the VM's metadata)

```
Sat May 25 19:42:08 2019 [Z0][VMM][E]: Error live migrating VM: Internal error Argument list too long - /var/lib/one/remotes/tm/storpool/premigrate vs06 vs05 /var/lib/one//datastores/0/201 201 0 PFZNP...
```
**To Reproduce**
* Instantiate a VM with 100 NIC Alias IPs (my test case)
* Try to live migrate the vm

**Expected behavior**
The live migrate should succeed

**Details**
 - Affected Component: [TransferManager]
 - Hypervisor: [KVM]
 - Version: [5.x]

**Additional context**
I'll push a patch that solves the issue but refactoring of all TM_MAD's is needed...

And will brake the addons if they are not updated, though.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

xorel:
PRs:
https://github.com/OpenNebula/one/pull/3583
https://github.com/OpenNebula/docs/pull/717

PR #3374, opened 2019-05-27
B #3373: pass VM template via STDIN

atodorov-storpool:
for _tm/*/{pre,post,fail}migrate_

### DIFF
--- a/src/datastore_mad/remotes/xpath.rb
+++ b/src/datastore_mad/remotes/xpath.rb
@@ -26,7 +26,8 @@ require 'pp'
 
 opts = opts = GetoptLong.new(
     [ '--stdin',   '-s', GetoptLong::NO_ARGUMENT ],
-    [ '--base64',  '-b', GetoptLong::REQUIRED_ARGUMENT ]
+    [ '--base64',  '-b', GetoptLong::REQUIRED_ARGUMENT ],
+    [ '--file',  '-f', GetoptLong::REQUIRED_ARGUMENT ]
 )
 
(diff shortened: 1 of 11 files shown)


xorel:
Hello @atodorov-storpool 
I modified it so the action scripts (tm/*/{pre,post,fail}migrate) receives VM template on stdin. 
Do you think you can test it? 
https://github.com/OpenNebula/one/commit/690dfb09f4398c14f010bf2cc9cb1c54c1925c55
thanks

atodorov-storpool:
Hi @xorel,

I think that the template should be stored in a temporary file that should be red by xpath_rb

Please take a look at https://github.com/OpenNebula/addon-storpool/blob/master/tm/storpool/premigrate#L47

And the patched xpath.rb :
https://github.com/OpenNebula/addon-storpool/blob/master/datastore/xpath-sp.rb
(the patch is not perfect, it just works for me :D )

xorel:
I don't think it's necessary. The error (``Argument list too long``) is only hit when the parameter is passed as a positional argument but still could be stored in a variable and passed through a pipe via stdin like here: 
https://github.com/OpenNebula/one/commit/690dfb09f4398c14f010bf2cc9cb1c54c1925c55#diff-606e289f74e14c1c2e1c35de93119c31R127

atodorov-storpool:
Fair enough :)

Can't spot other issues

xorel:
Created https://github.com/OpenNebula/one/pull/3583 instead

Issue #3269, opened 2019-04-24
OOM killer: onedb purge-done eat too much RAM


baby-gnu:
**Description**

After upgrading to 5.8.1, I wanted to shrink a little my MySQL database (12GB on disk for 5.6GB uncompressed SQL file).

I tried the `onedb purge-done` command but it was killed by the kernel, causing an Out Of Memory Killer run.

**To Reproduce**

We have 211 768 VMs in our MySQL database since 2014, I just ran `onedb purge-done`.

**Expected behavior**

The command should not eat all the memory.

**Details**
 - Affected Component: Database
 - Hypervisor: KVM
 - Version: 5.8.1

**Additional context**

@xorel told me to use the `--start` and `--end` option, so I did a purge 3 months by 3 months. It take around 6GB of RAM on our 12GB frontend to run.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

baby-gnu:
I did not have the same issue with `onedb purge-history`.

al3xhh:
PRs to merge in master and 5.8:

Code: https://github.com/OpenNebula/one/pull/3376
Docs: https://github.com/OpenNebula/docs/pull/635

al3xhh:
Hello @baby-gnu 

I have made an improvement on this, could you please try it and let me know if it works for you?

Thanks!

PR #3376, opened 2019-05-28
B #3269: onedb purge-done eat too much RAM

al3xhh:
* Process done VMs in pages, instead of all together
    * Add new method to iterate over pages
    * Add new CLI paramter to specify pages size

### DIFF
--- a/src/onedb/onedb_live.rb
+++ b/src/onedb/onedb_live.rb
@@ -5,7 +5,7 @@ require 'base64'
 class OneDBLive
 
     EDITOR_PATH = '/bin/vi'
-    PAGES       = 100
+    PAGES       = 0
 
     def initialize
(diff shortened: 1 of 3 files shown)


src/onedb/onedb_live.rb
```diff
@@ -96,6 +97,19 @@ def select(table, where)
         [row].flatten.compact
     end
 
+    def count(table, where)
+        sql = "SELECT COUNT(*) FROM #{table} WHERE #{where}"
+        res = db_query(sql, 'Error querying databse')
```
tinova: small typo: database

al3xhh: Ups, that function shouldn't be there, thanks @tinova for noticing that!

Issue #3263, opened 2019-04-23
Simplify and make the FSCK network code more clear


al3xhh:
**Description**
Divide the code into functions so it's easier to read and also fix some bugs.

**Interface Changes**
Network FSCK.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
PRs to merge in master and 5.8:

Code: https://github.com/OpenNebula/one/pull/3275
Docs: https://github.com/OpenNebula/docs/pull/603
Tests: https://github.com/OpenNebula/development/pull/613

PR #3391, opened 2019-05-31
F #3263: fix bug with mac

al3xhh:


al3xhh:
Applies to master and 5.8

Issue #1710, opened 2018-02-03
onedb fsck not count leases from reserverd networks


kvaps:
# Bug Report
## Version of OpenNebula

- [ ] 5.2.2
- [ ] 5.4.0
- [ ] 5.4.1
- [ ] 5.4.2
- [ ] 5.4.3
- [ ] 5.4.4
- [ ] 5.4.5
- [X] 5.4.6
- [ ] 5.4.7
- [ ] Development build

## Component

- [ ] Authorization (LDAP, x509 certs...)
- [x] Command Line Interface (CLI)
- [ ] Contextualization
- [ ] Documentation
- [ ] Federation and HA
- [ ] Host, Clusters and Monitorization
- [ ] KVM
- [ ] Networking
- [ ] Orchestration (OpenNebula Flow)
- [ ] Packages
- [ ] Scheduler
- [x] Storage & Images
- [ ] Sunstone
- [x] Upgrades
- [ ] User, Groups, VDCs and ACL
- [ ] vCenter

## Description

I have virtual network with 23, and network reservation for each user from it.

### Expected Behavior

onedb fsck checks network leases in reserved network too.

### Actual Behavior

onedb fsck break network quotas set it to "0":

```
User 350 quotas: VNet 23        LEASES_USED has 1       is      0
User 351 quotas: VNet 23        LEASES_USED has 1       is      0
User 352 quotas: VNet 23        LEASES_USED has 1       is      0
User 353 quotas: VNet 23        LEASES_USED has 1       is      0
User 355 quotas: VNet 23        LEASES_USED has 1       is      0
User 356 quotas: VNet 23        LEASES_USED has 1       is      0
Group 150 quotas: VNet 23       LEASES_USED has 1       is      0
Group 151 quotas: VNet 23       LEASES_USED has 1       is      0
Group 152 quotas: VNet 23       LEASES_USED has 1       is      0
Group 153 quotas: VNet 23       LEASES_USED has 1       is      0
Group 155 quotas: VNet 23       LEASES_USED has 1       is      0
Group 156 quotas: VNet 23       LEASES_USED has 1       is      0
```

## How to reproduce

* Create network reservation
* Run onedb fsck

kvaps:
Issue still there on 5.6.1

kvaps:
Patch for disable this check:

```diff
--- a/usr/lib/one/ruby/onedb/fsck/quotas.rb
+++ b/usr/lib/one/ruby/onedb/fsck/quotas.rb
[...]

PR #3405, opened 2019-06-05
B #1710: fix bug when parent_id is empty

al3xhh:
Applies to 5.8 and master.

### DIFF
--- a/src/onedb/fsck/quotas.rb
+++ b/src/onedb/fsck/quotas.rb
@@ -474,11 +474,13 @@ module OneDBFsck
             vnet_doc.xpath('VNET/AR_POOL').each do |ar|
                 parent_id = ar.xpath('AR/PARENT_NETWORK_AR_ID')
                 parent_id = parent_id.text unless parent_id.nil?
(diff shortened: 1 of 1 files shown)


src/onedb/fsck/quotas.rb
```diff
@@ -474,11 +474,13 @@ def calculate_vnet_quotas(doc, queries, resource)
             vnet_doc.xpath('VNET/AR_POOL').each do |ar|
                 parent_id = ar.xpath('AR/PARENT_NETWORK_AR_ID')
                 parent_id = parent_id.text unless parent_id.nil?
-                size      = ar.xpath('AR/SIZE').text.to_i
 
-                if !parent_id.nil?
-                    vnet_usage[parent_id] = 0 if vnet_usage[parent_id].nil?
-                    vnet_usage[parent_id] += size
+                next if parent_id.nil? || parent_id.empty?
```
screenlock: Is it correct ? Pass if its nil or its empty, better not (nil or empty) 
next if !(parent_id.nil? || parent_id.empty?)
i think

Issue #3406, opened 2019-06-05
Enhance LXD cleanup


OpenNebulaSupport:
**Description**
LXD error code handling can fail and [leave unmapped disks](https://github.com/OpenNebula/one/issues/3389) and stopped containers.

**Expected behavior**
Containers nor mapped disks shouldn't be hanging in the node after failed operations.

**Details**
 - Hypervisor: LXD
 - Version: 5.8.x

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

Issue #3389, opened 2019-05-30
LXD fails to cleanup context CD-ROM mount


vholer:
**Description**
Sometimes it may happen the LXD container disk mount for the contextualization CD-ROM isn't unmounted properly. Final rm on the VM datastore directory fails, and leaves the VM in EPILOG_FAILURE state.

Only this error can be found in the VM log:

```
Wed May 29 19:30:03 2019 [Z0][TM][I]: Command execution failed (exit code: 1): /var/lib/one/remotes/tm/qcow2/delete ubuntu1804-context-lxd-9702e-0.test:/var/lib/one//datastores/0/98 98 0
Wed May 29 19:30:03 2019 [Z0][TM][I]: delete: Deleting /var/lib/one/datastores/0/98
Wed May 29 19:30:03 2019 [Z0][TM][E]: delete: Command "[ -e "/var/lib/one/datastores/0/98" ] || exit 0
Wed May 29 19:30:03 2019 [Z0][TM][I]: 
Wed May 29 19:30:03 2019 [Z0][TM][I]: times=10
Wed May 29 19:30:03 2019 [Z0][TM][I]: function="rm -rf /var/lib/one/datastores/0/98"
Wed May 29 19:30:03 2019 [Z0][TM][I]: 
Wed May 29 19:30:03 2019 [Z0][TM][I]: count=1
Wed May 29 19:30:03 2019 [Z0][TM][I]: 
Wed May 29 19:30:03 2019 [Z0][TM][I]: ret=$($function)
Wed May 29 19:30:03 2019 [Z0][TM][I]: error=$?
Wed May 29 19:30:03 2019 [Z0][TM][I]: 
Wed May 29 19:30:03 2019 [Z0][TM][I]: while [ $count -lt $times -a "$error" != "0" ]; do
Wed May 29 19:30:03 2019 [Z0][TM][I]: sleep 1
Wed May 29 19:30:03 2019 [Z0][TM][I]: count=$(( $count + 1 ))
Wed May 29 19:30:03 2019 [Z0][TM][I]: ret=$($function)
Wed May 29 19:30:03 2019 [Z0][TM][I]: error=$?
Wed May 29 19:30:03 2019 [Z0][TM][I]: done
Wed May 29 19:30:03 2019 [Z0][TM][I]: 
[...]

PR #3410, opened 2019-06-07
F #3406: Fix nil mappers. Centralize disk filter.

dann1:
Basically we have our mapper generation prone to being nil by not matching the supported mappers.

- Also simplified some logic centralization

1st of probably more PRs related to #3406, there could be more situation leading to map issues like #3389.


dann1:
- The container will now boot and avoid mapping the invalid disks
- disks being hotplugged won't be skipped and the operation will fail without the disk being mapped on the host
- if there is an error when setting up the container storage (skipping the unsupported disks aren't errors), the container will be deleted only if the unmap operation is successfully executed.

Issue #3463, opened 2019-06-28
Error when try delete image in vCenter. Using nfs not delete the image on disk.


carloshm91:
**Description**
When you execute the command:

oneimage delete 'name_of_image'

if you have the image inside a nfs datastore got this error:

Error removing image from datastore. Manually remove image source X to completely delete the image.

**To Reproduce**
1) Import a image to nfs datastore
2) Instantiate the image
3) Delete the VM instantiated 
4) Delete image

**Expected behavior**
Delete the image from disk

**Details**
 - Affected Component: [Storage]
 - Hypervisor: [vCenter]
 - Version: [5.8.*, master]

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

tinova:
PR #3464

PR #3464, opened 2019-06-28
B #3463: fix error when delete image on vcenter using nfs

carloshm91:
to merge with master


tinova:
also to one-5.8

Issue #3490, opened 2019-07-08
[GOCA] Login method sends wrong parameter


trenton42:
**Description**
When calling the `Login` method on a user, GOCA sends `user.ID` as the first parameter, but the server expects `user.Name`.

**To Reproduce**
1. Get a user object
2. Make the `Login` call with any parameters.
3. The server will respond with the error:
```
Error: GOCA client error [RESPONSE_XMLRPC_FAULT]: server response: error: "Parameter that is supposed to be a string is not" code: -501
```

**Expected behavior**
No error is returned, and a login token is created

**Details**
 - Affected Component: goca
 - Hypervisor: any
 - Version: all versions of goca

**Additional context**
See API docs: http://docs.opennebula.org/5.8/integration/system_interfaces/api.html#one-user-login

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #3491, opened 2019-07-08
B #3490 Fix wrong parameter for one.user.login

trenton42:
- `one.user.login` expects username, not id
- Clarified doc string
- For #3490


trenton42:
Thank you @rsmontero!

Issue #3520, opened 2019-07-15
[GOCA] Template schema lacks NIC field


eripa:
**Description**
The GOCA template schema in https://github.com/OpenNebula/one/blob/master/src/oca/go/src/goca/schemas/template/template.go lacks the `NIC` field, and `NIC` doesn't properly work in the Dynamically unmatched tags.

**To Reproduce**
Create a Template, try to access any NIC property using the GOCA library.

**Expected behavior**
The GOCA library should provide means to access the NIC properties using the Template schema.

**Additional context**
I have a pull request ready to go that I've tested successfully on my machine. I'll push that and reference this issue.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

treywelsh:
It seems that I did some mistakes :)

Just to notify you that there is still some pending changes.
Among others, there is a bug in template dynamic parts, and overall, they could be improved. 
I already have some code, I hope I could propose a PR soon to improve this part.

eripa:
hey @treywelsh, any chance there's some progress on the template rework? I've looked at your fork's branch and there seem to be some good things coming.

treywelsh:
I finally propose my changes in a new pull request.
I'm testing it, but as it remains a lot of work to do on the tests, I could miss a lot of things anyway.
Feel free to comment/test it I'm open to discuss :)

Issue: https://github.com/OpenNebula/one/issues/3219
PR: https://github.com/OpenNebula/one/pull/3837

PR #3521, opened 2019-07-15
B #3520: Add missing NIC field to Template schema in GOCA

eripa:
Solves Issue #3520


eripa:
Any chance we can merge this trivial change?

PR #3543, opened 2019-07-23
M #: fix problem with white spaces in onegate parameter

al3xhh:
Applies to 5.8 and master.


al3xhh:
Applies to 5.8 and master.

al3xhh:
I tested and it doesn't break, but I will test it again.

If we have that, it behaves in strange way, the best way is just have one ".

Issue #3556, opened 2019-07-25
IP spoofing filter doesn't block traffic between VMs on the same host


vmoyankov:
**Description**
On the bridged network with enabled IP spoofing filter, when the destination interface is a VM and input rules permit the traffic, packet is accepted with `-j ACCEPT` and may skip IP spoofing filters defined later in the chain `opennebula`. This allows traffic with spoofed IP and MAC address between VMs on the same host.

**To Reproduce**

 * Create a VNet of type "Bridged & Security groups" (VN_MAD = fw).
Enable MAC spoofing filter
Enable IP spoofing filter
Add IP Address range.

 * Create two VMs, each with one interface in this VNet. Start the VMs. Let assume ONE assigned IP addresses 10.2.86.1 and 10.2.86.2 to the VMs.

 * In each VM add manually additional IP address from the same range (e.g. 10.2.86.11 and 10.2.86.12):
vm1# ip address add 10.2.86.11/24 dev eth0
vm2# ip address add 10.2.86.21/24 dev eth0

 * Ping VMs from the spoofed source address:
vm1# ping -I 10.2.86.11 10.2.86.2
and
vm2# ping -I 10.2.86.21 10.2.86.1

Depending on the order of filters in the `opennebula` chain, one of the ping will succeed.

**Expected behavior**
All packets with spoofed source IP shall be dropped.

**Details**
 - Affected Component: [network]
 - Hypervisor: [KVM]
 - Version: [5.8.4]

**Additional context**
Pull request with a fix will be proposed.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #3558, opened 2019-07-25
B #3556: Fix IP spoofing filters

vmoyankov:
Signed-off-by: Venko Moyankov <vmoyankov@gmail.com>


rsmontero:
Thanks!. Yes this is in fact the approach used in the other rules

PR #3570, opened 2019-08-02
F 3440 Refactor and corrections

mouyaq:
Refactor of host.rb related to nsx and error corrections.
Updated incorrect install.sh


tinova:
not merging at the momento due to rubocop warning

PR #3573, opened 2019-08-06
fix condition for calculate quotas for reserved networks

telecast-git:
It’s seems additional quotas on not exist resourses is working from this condition:

fsck/quotas.rb

    475                 parent_id = ar.xpath('AR/PARENT_NETWORK_AR_ID')
    476                 parent_id = parent_id.text unless parent_id.nil?
    477 
    478                 next if parent_id.nil? || parent_id.empty?
    479 
    480                 vnet_usage[parent_id] = 0 if vnet_usage[parent_id].nil?
    481 
    482                 ar.xpath('AR/SIZE').each do |size|
    483                     vnet_usage[parent_id] += size.text.to_i
In this case if “PARENT_NETWORK_AR_ID” not exist we will be add VALUE to empty KEY, i.e. instead PARENT_NETWORK_AR_ID = [SIZE] we have __ = [SIZE].
May be will more logical if this will so:

 478                 next if !parent_id.nil? || !parent_id.empty?

If we meet empty or nil KEY PARENT_NETWORK_AR_ID, will missed it and won’t added additional quotas on not exist resources.

### DIFF
--- a/src/onedb/fsck/quotas.rb
+++ b/src/onedb/fsck/quotas.rb
@@ -475,7 +475,7 @@ module OneDBFsck
                 parent_id = ar.xpath('AR/PARENT_NETWORK_AR_ID')
                 parent_id = parent_id.text unless parent_id.nil?
 
-                next if parent_id.nil? || parent_id.empty?
+                next if !parent_id.nil? || !parent_id.empty?
 
                 vnet_usage[parent_id] = 0 if vnet_usage[parent_id].nil?
 


screenlock:
possible duplicate request #3405

screenlock:
Oh sorry. This trouble comes from syntax. Keyword "next" skips current part if condition is true. Thats mean that this statement is true.

al3xhh:
Hello @telecast-git 

I have checked your PR and it seems a bit strange for me. I have done some tests and with your changes it doesn't work, the condition you added will skip all reserved networks.

> In this case if “PARENT_NETWORK_AR_ID” not exist we will be add VALUE to empty KEY, i.e. instead PARENT_NETWORK_AR_ID = [SIZE] we have __ = [SIZE].

With the current condition we avoid this, with your change we broke it. So please try to elaborate more your issue and I can help you fixing it.

al3xhh:
Closing the PR, you can open a new one if it's necessary.

Issue #3563, opened 2019-07-26
Error importing vm due to guest.net.ipConfig property


mouyaq:
**Description**
Error importing a VM with guest.net.ipConfig property unset.

**To Reproduce**
Try to import a VM which object in vcenter has the attribute guest.net.ipConfig unset.

**Expected behavior**
This must be an optional value retrieved from VMware Tools, so if no info is getting should be ommited

**Details**
 - Affected Component: vCenter driver
 - Hypervisor: ESXi
 - Version: all

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

PR #3609, opened 2019-08-22
B #3563: Fix VM import

mouyaq:
Fix the problem when trying to import a VM with
guest.net.ipConfig attribute unset.


mouyaq:
Merge on Master and one-5.8

Issue #3597, opened 2019-08-16
noVNC is not working for VMs with password enabled


kvaps:
**Description**
After upgrade to 5.8.4 I faced with the issue when most of VMs have no working VNC. Usually it is stacks on this screen without any errors:
![image](https://user-images.githubusercontent.com/7556217/63196383-764d6100-c075-11e9-968d-e335d040fb4d.png)

But after few minutes, it shows:
![image](https://user-images.githubusercontent.com/7556217/63196557-d2b08080-c075-11e9-9bad-f8d4db8ac044.png)

and error in browser's console:
```
Failed when connecting: Connection closed (code: 1006) rfb.js:677
    _fail rfb.js:677
    RFB rfb.js:277
    onclose websock.js:227
```

**To Reproduce**

- Create VM with 
  ```
  GRAPHICS = [
    LISTEN = "0.0.0.0",
    RANDOM_PASSWD = "YES",
    TYPE = "VNC" ]
  ```
- Try to connect via VNC

**Expected behavior**
Connected VNC

**Details**
 - Affected Component: Sunstone
 - Hypervisor: KVM
 - Version: 5.8.4

**Additional context**

After rollback to 5.8.1 vnc is working again

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

alitvak69:
I have noticed some slowness, whence a black bar in the beginning.   However, it loads the console in a few seconds.

kvaps:
Looks like novnc client problem, I have sunstone and novnc in separated containers.

So I can confirm that problem not in novnc-server but in client, I've tested few scenarios:

* Sunstone 5.8.1 and novnc server from 5.8.4 - working fine
* Sunstone 5.8.4 and novnc server from 5.8.4 - can't connect to some VMs

I also made dump of packets on the host with the both VMs to see what is actually wrong there:

Left side - bad case, Right side - good case
![image](https://user-images.githubusercontent.com/7556217/63212614-7f870e00-c107-11e9-9820-4ec7a6284eea.png)

What did I found, actually nothing, just see that after 10th packets novnc-server does not sends any packets.

kvaps:
Got it! Problem occurs only for VMs with enabled password authentication for VNC

[...]

PR #3612, opened 2019-08-23
B #3597: Fix send credentials noVNC

tryfunc:
Fixed sending noVNC credentials with password enabled.

### DIFF
--- a/src/sunstone/public/app/console/vnc.js
+++ b/src/sunstone/public/app/console/vnc.js
@@ -104,6 +104,7 @@ define(function(require) {
   var proxy_host = window.location.hostname;
   var proxy_port = Config.vncProxyPort;
   var token = getQueryVariable("token");
+  var pw = getQueryVariable("password");
 
   if (window.location.protocol === "https:") {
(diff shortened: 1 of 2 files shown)


tryfunc:
@jloboescalona2 Why make a separate PR for the same changes? PR #3651 

In addition, to add additional parameters in the future, it would be more logical to do otherwise:

```
var rfbConfig = password? { "credentials": { "password": password } } : {};
```

```
var rfbConfig = { "credentials": { "password": password || null } };
```

`credentials` - An `Object` specifying the credentials to provide to the server when authenticating. The following credentials are possible:

| name         | type        | description
| ------------ | ----------- | -----------
| `"username"` | `DOMString` | The user that authenticates
| `"password"` | `DOMString` | Password for the user
| `"target"`   | `DOMString` | Target machine or session

jloboescalona2:
hi, @tryfunc. Sorry for the inconvenience, but in your PR not valid, the case that does not send password, in that case the RFB configuration should remain the same call to NoVNV, thanks for your contribution

Issue #3631, opened 2019-09-01
Incorrect DEV_PREFIX when create image using command "oneimage create"


carloshm91:
**Description**
When you use the command "oneimage create" using for the input image a qcow2 image and put the output to a vcenter datastore, you will get: 

DEV_PREFIX	hd	   
DRIVER	        qcow2

**To Reproduce**
You need have a instance of ONE running with a vcenter imported.
Then you can execute the next command

> oneimage create -d 'nfs(IMG)' --type OS --name 'Image Name' --path 'http://server/where/is/image/image.qcow2' --driver qcow2

**Expected behavior**
In the image template
DEV_PREFIX	sd

**Details**
 - Hypervisor: [vCenter]
 - Version: [one-5.8 and master]

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

tinova:
needs change in oned.conf to inherit DEV_PREFIX from DS
also needs documentation in vcenter section

carloshm91:
[DOC] 
* https://github.com/OpenNebula/docs/pull/766
[ONE] 
* https://github.com/OpenNebula/one/pull/3632

PR #3632, opened 2019-09-01
B #3631: dev_prefix on create image

carloshm91:
-> master
-> one-5.8


tinova:
only to master, cannot change conf files in maintenance releases

Issue #3440, opened 2019-06-18
Add NSX-V support to OpenNebula's vCenter driver


mouyaq:
**Description**
These are the NSX-V features we expect to work on vcenter driver.

- Logical Switches

- Security Groups

- Distributed Firewall

- Edge Gateway Services

**Use case**
In a first instance:
Allows working with logical switches, to attach, detach, create, etc
Use, create, delete, security groups and distributed firewall.
In a second instance:
EDGE services: FW, NAT, VPN, etc.

**Interface Changes**
New cli commands for NSX interface and sunstone menus. 

**Additional Context**

## Progress Status
- [x] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #3636, opened 2019-09-02
F #3440: Fix monitoring loop

mouyaq:
Fix a monitoring loop that occurs when NSX Manager is not
reachable. An SSH exception is created an created a personalized
message that break the format of the attribute NSX_STATUS.
That message has been changed to a static message.

Also fix a typo when launching an Standard Error.


tinova:
only in master

Issue #3042, opened 2019-03-06
Packet Public Networking


vholer:
**Description**
Port the Packet IPAM prototype from the DDC3, update for the current ONE code, library locations, etc. It should support the general management of the public IPv4 ranges on Packet (allocation and release, note: mapping on the hosts / VMs will be handled by different component):
https://github.com/OpenNebula/ddc/tree/master/src/ipamm_mad/remotes/packet

We need a dedicated virtual network driver for Packet, which assigns the NIC aliased public IP address to the Packet physical host (as part of pre action), and configures the S/DNAT translation between the VM public IP address and VM private IP address (as part of post action). This should leverage the new external NIC aliases functionality = public IP isn't configured inside the VM, but the traffic is only "routed" to the VM private IP.

**Use case**
Manage Packet public IPv4 ranges from the OpenNebula.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

cloudweavers:
Link returns 404... Is it a private repo?

vholer:
Yes, it's private repo.

al3xhh:
The code has been ported and linted, I have also implemented some tests.

vholer:
Reopening, we should rename few parameters:

- `PROJECT_ID` to `PACKET_PROJECT`
- `TYPE` to `PACKET_IP_TYPE`

(low priority, not blocking anything)

vholer:
Also, about the PACKET_IP_TYPE, I wouldn't be fixed on the IPv4 only here:

```
"#{ar_data.xpath('//AR/TYPE').text.downcase}_ipv4"
```

and in the AR template below. We can expect the number of supported types to grow, so for now probably **public_ipv4** and **global_ipv4** (TODO: request to enter the full keyword) are supported over API. But, over portal I can already request even IPv6 ranges.

I can't tell which all types are now valid, as the list of options is not documented:
https://www.packet.com/developers/api/#requesting-ip-reservations

al3xhh:
Code: https://github.com/OpenNebula/one/pull/3519
Docs: https://github.com/OpenNebula/docs/pull/701
Tests: https://github.com/OpenNebula/development/pull/679

vholer:
[...]

PR #3639, opened 2019-09-03
F #3042: Respect GEMS_LOCATION, use STDIN for data

vholer:
Adapts code to
- new dedicated gems locations
- new way of passing vnet drivers data over stdin instead of argument

To be merged just into **master**.


vholer:
image

Issue #3241, opened 2019-04-17
ipv6 address is being shown for a part of a second and then replaced by "--" 


jloboescalona2:
**Description**
the ipv6 address is being shown for a part of a second and then replaced by "--" -- I assume it is a width / javascript thing that makes it "invisible"

**Details**
 - Affected Component: [e.g. Sunstone, Scheduler or Storage]
 - Hypervisor: [e.g. KVM]
 - Version: [e.g. 5.4.12 or development]

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [x] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

jloboescalona2:
PR links:
-Docs: https://github.com/OpenNebula/docs/pull/740
-One: https://github.com/OpenNebula/one/pull/3659

PR #3659, opened 2019-09-05
B #3241: show ipv6 address

jloboescalona2:
this PR apply:
-master
-one-5.8-maintenance

### DIFF
--- a/src/sunstone/public/app/tabs/vms-tab/panels/network.js
+++ b/src/sunstone/public/app/tabs/vms-tab/panels/network.js
@@ -211,17 +211,25 @@ define(function(require) {
     return html;
   }
 
-  function _ipTr(nic, attr){
+  function _ipTr(nic, attr, attr2){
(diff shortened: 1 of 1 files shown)


src/sunstone/public/app/tabs/vms-tab/panels/network.js
```diff
@@ -211,17 +211,25 @@ define(function(require) {
     return html;
   }
 
-  function _ipTr(nic, attr){
+  function _ipTr(nic, attr, attr2){
```
tinova: would be clear to receive an array and iterate over it

feldsam:
Hello @jloboescalona2 and @tinova, this commit breaks dual stack, ipv4 with ipv6 setup. On network tab I can see only ipv6 addresses. in place where was IP4 is IP6_LINK now.
Screenshot 2019-10-27 at 00 31 38

Issue #3613, opened 2019-08-25
brctl is obsolete and should be replaced in code


feldsam:
**Description**
`brctl` command is obsolete and `bridge-utils` package is not in REHL8, Oracle Linux 8 and in upcoming CentOS 8. There is also no package in EPEL-8.0 release.

`brctl` calls in VNM drivers should be replaced by `ip link` commands. 

**Details**
 - Affected Component: Network
 - Hypervisor: KVM, LXD
 - Version: All

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

vholer:
Also related to brctl #3522

xorel:
PRs:
* one https://github.com/OpenNebula/one/pull/3654 to master
* docs https://github.com/OpenNebula/docs/pull/736 to master

feldsam:
just note, bridge-utils dep have to be removed from packing spec files in this repo https://github.com/OpenNebula/packages

PR #3661, opened 2019-09-05
F #3613: Fixed vlan driver on el8 systems

feldsam:
Command `ip link show master br_name` list attached vlan devices with @master_dev on el8 systems. For example `team0.404@team0`. This commit fixes parsing bridges and attached devices.


xorel:
Thanks!

Issue #3380, opened 2019-05-29
Redesign hook system to add new features


christian7007:
**Description**
The goal of this new feature is to extend the current hook mechanism to address two common requests:
   - Better feedback on hook execution. Currently it is a fire-and-forget without any feedback (limited to oned.log) and no retry capabilities.
   - Being able to hook on API calls. Currently hooks are mainly based on state changes.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

feldsam:
nice to have hooks executed on device create and delete events

PR #3735, opened 2019-09-19
F #3380: Add auto-register vcenter hooks

mouyaq:
vCenter hooks located at:

 <ONE_LOCATION>/remotes/hooks/vcenter/templates

are registered automatically when import a vcenter cluster
with the command "onevcenter"

### DIFF
--- a/src/vmm_mad/remotes/lib/vcenter_driver/vcenter_importer.rb
+++ b/src/vmm_mad/remotes/lib/vcenter_driver/vcenter_importer.rb
@@ -15,6 +15,18 @@
 #--------------------------------------------------------------------------- #
 
 module VCenterDriver
+
+    ONE_LOCATION = ENV['ONE_LOCATION'] unless defined?(ONE_LOCATION)
+
+    if !ONE_LOCATION
+        RUBY_LIB_LOCATION = '/usr/lib/one/ruby' unless defined?(RUBY_LIB_LOCATION)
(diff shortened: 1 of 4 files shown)


src/vmm_mad/remotes/lib/vcenter_driver/vcenter_importer.rb
```diff
@@ -521,5 +523,21 @@ def self.import_clusters(con_ops, options)
 
         end
 
+        def self.register_hooks
+            if !ONE_LOCATION
+                hooks_path = '/var/lib/one/remotes/hooks/vcenter/templates'
+            else
+                hooks_path = ONE_LOCATION + '/remotes/hooks/vcenter/templates'
+            end
```
tinova: this should be moved to the beginning of the file (as in the other driver files)

src/vmm_mad/remotes/lib/vcenter_driver/vcenter_importer.rb
```diff
@@ -521,5 +523,21 @@ def self.import_clusters(con_ops, options)
...
+            client = OpenNebula::Client.new
+            hook = OpenNebula::Hook.new(OpenNebula::Hook.build_xml, client)
+            hook_pool = OpenNebula::HookPool.new(client)
+            hook_pool.info
```
tinova: check rc for errors

PR #3738, opened 2019-09-19
F #3738: Secure cookie flag if scheme is https

xBytez:
Build in functionality so Sunstone marks cookies set as Secure to ensure only secure servers can access the session cookie.

I've also made [a PR to OpenNebula/docs](https://github.com/OpenNebula/docs/pull/771) to document this change.

PS: It'd be nice to have a ISSUE_TEMPLATE so I know what GitCop expects so I don't make this mistake of writing PRs this way.

### DIFF
--- a/src/sunstone/sunstone-server.rb
+++ b/src/sunstone/sunstone-server.rb
@@ -411,6 +411,12 @@ helpers do
 
         # end user options
 
+        # secure cookies
+        if request.scheme == 'https'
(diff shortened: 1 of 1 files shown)


tinova:
Hi, thanks a lot for your quality contribution!

About GitCop, you can check the guidelines here: https://github.com/OpenNebula/one/blob/master/share/doc/dev/COMMIT_MESSAGES.md

PR #3755, opened 2019-09-23
M #-: fix bug in onedb purge done

al3xhh:
Fix bug when there are no consecutive VMs, so the get_page
    function didn't return the pool.


al3xhh:
Applies to 5.8 and master.

Issue #3567, opened 2019-07-30
Missing parameters when import a vcenter network


mouyaq:
**Description**
After import a vCenter network, some mandatory parameters aren't imported, like switch name, and host id.

**To Reproduce**
Import a vCenter network, and try to update the network config. 
Switch name and OpenNebula's Host ID will be empty

**Expected behavior**
Those attributes must be set at network import or if those attributes are not mandatory change them.

**Details**
 - Affected Component: vCenter driver
 - Hypervisor: ESXi
 - Version: all

**Additional context**

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

carloshm91:
PR -> https://github.com/OpenNebula/one/pull/3838

PR #3838, opened 2019-10-13
B #3567: add switch name in DPG

carloshm91:
-> one-5.10
-> master


src/vmm_mad/remotes/lib/vcenter_driver/network.rb
```diff
@@ -185,6 +186,8 @@ def self.to_one(network_import_name, network_name, network_ref, network_type,
 
         template += "VCENTER_TEMPLATE_REF=\"#{template_ref}\"\n" if template_ref
 
+        template += "VCENTER_TEMPLATE_REF=\"#{sw_name}\"\n" if sw_name
```
tinova: needs to be a different attribute

PR #3857, opened 2019-10-17
F #4556: Added option to conf VM logs location

feldsam:
This is usefull in HA Raft setup to easily share VM logs between HA nodes

### DIFF
--- a/share/etc/oned.conf
+++ b/share/etc/oned.conf
@@ -59,12 +59,14 @@
 #      syslog    to use the syslog facilities
 #      std       to use the default log stream (stderr) to use with systemd
 #   debug_level: 0 = ERROR, 1 = WARNING, 2 = INFO, 3 = DEBUG
+#   use_vms_location: defines if store VM logs in VMS_LOCATION
 #
 #*******************************************************************************
 
 LOG = [
   SYSTEM      = "file",
-  DEBUG_LEVEL = 3
+  DEBUG_LEVEL = 3,
+  USE_VMS_LOCATION = "NO"
 ]
 
 #MANAGER_TIMER = 15
--- a/src/nebula/Nebula.cc
+++ b/src/nebula/Nebula.cc
@@ -1290,8 +1290,16 @@ void Nebula::get_ds_location(string& dsloc)
 string Nebula::get_vm_log_filename(int oid)
 {
     ostringstream oss;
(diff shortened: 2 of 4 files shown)


feldsam:
hello, for some reason this doesnt work, would somebody point out why?

al3xhh:
Hello @feldsam the commit message should be "F/B #<issue_number>: <commit_message>"

feldsam:
Hi @al3xhh I know, but I was meaning that my code changes doesnt work :) and oned still logs to /var/log/one/vmid.log

feldsam:
Hello, it is working in my env, now. What you think about this implementation? If this will be accepted or we agree on some other solution, then I plan to write blog post about setuping HA using RAFT and syncing logs using lsyncd.

I tested this with lsyncd and updated raft hook script, whichs stop/start lsyncd on nodes. Looks like that it is sufficient to just sync /var/lib/one/vms/ directory, when shared storage is used.

feldsam:
Should I create issue for this?

feldsam:
Example `lsyncd` config on `one1` frontend node:

```
targets = {
    'oneadmin@one1.feldcloud.net:/var/lib/one/vms/',
    'oneadmin@one2.feldcloud.net:/var/lib/one/vms/',
    'oneadmin@one3.feldcloud.net:/var/lib/one/vms/',
}

for _, target in ipairs( targets )
do
    if not string.match(target, "one1.feldcloud.net") then
        sync{
          default.rsync,
          source = "/var/lib/one/vms/",
          target = target,
          rsync = {
            archive = true,
            rsh = "/usr/bin/ssh -l oneadmin -i /var/lib/one/.ssh/id_rsa -o StrictHostKeyChecking=no"
          },
          delay = 1
        }
    end
end
```

In RAFT hook `lsyncd` service is started on the leader and stopped on followers.

rsmontero:
Could you configure lsyncd to just sync [0-9]*.log files in /var/log/one/?

feldsam:
Hi, `lsyncd` uses `rsync`, which doesn't speak regex. In VMS location are also other files, and this seemed the most natural solution to me :)

rsmontero (approved):

rsmontero:
Merged as per b9e2dcc4ebf15cd489ecfc48a1ca61d3f6844e17

PR #3910, opened 2019-11-05
F #3380: Generate man pages for onehook command

christian7007:


vholer:
Duplicate entry, `onehook` is already there. We have manual page packaged

```
root@debian10-kvm-nfs-ovs-5-9-90-84af1-0:~# man onehook | head
ONEHOOK(1)          onehook(1) -- manages OpenNebula hooks          ONEHOOK(1)

NAME
       onehook - manages OpenNebula hooks

SYNOPSIS
       onehook command [args] [options]

OPTIONS
        --dry                     Just print the template
```

Issue #3990, opened 2019-11-25
NSX networks should be listed only if NSX_STATUS = OK


mouyaq:
**Description**
NSX networks are represented as network objects in vcenter, so when you list a network to import it is detected, but if you haven't introduced NSX credentials yet, you cannot import it.
In the network list should appear only importable networks.
the command onevcenter list_all could show all networks but list should only list importable ones.

**Use case**
Good to avoid confusions about what networks can be imported.

**Interface Changes**
vcenter driver, nsx driver

**Additional Context**
Add any other context or screenshots about the feature request here. Or any other alternative you have considered to addressed this new feature.

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

mouyaq:
Manually tested, waiting automatic testing results.

mouyaq:
PR to doc: https://github.com/OpenNebula/docs/pull/861

mouyaq:
Related with this PRs: 
https://github.com/OpenNebula/one/pull/4058
https://github.com/OpenNebula/one/pull/4059

PR #4016, opened 2019-12-03
B #3990: List NSX networks only when NSX_STATUS=OK

mouyaq:


tinova:
also cherry pick to one-5.10

Issue #4026, opened 2019-12-06
kvm VM_MAD should not fail when ejecting already ejected context cdrom


atodorov-storpool:
**Description**
OpenNebula should not fail updating/reconfiguring contextualization when the cdrom image is already ejected.

**Use case**
If updating the reconfigure action fail for some reason the VM will have the cdrom image ejected and on next reconfigure the prereconfigure script will fail.

**Interface Changes**
Check that contextualization CDROM is already ejected during prereconfigure.

**Additional Context**
This change will improve the OpenNebula stability on some corner cases.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #4028, opened 2019-12-06
F #4026: do not fail when CDROM already ejected

atodorov-storpool:


xorel:
LGTM

Issue #4076, opened 2020-01-04
Missing ACL_POOL in XSD schema


feldsam:
**Description**
ACL Pool is defined, but not included in XSD index. There is also wrong definition on most field, which are integers but in reality are hexadecimals

**Details**
 - Affected Component: Core
 - Hypervisor: All
 - Version: All

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
Also needs xsd tests

feldsam:
point me to dir where tests lives

rsmontero:
NP it was more a note for us when reviewing, and merging. THANKS for your contributions and feed back!

xorel:
tests (xsd + pyone) https://github.com/OpenNebula/development/pull/812

feldsam:
@rsmontero hello, is this possible to cherry-pick also to 5.10? Because the final version `Use xs:string for ACLs` is only in the master branch.

rsmontero:
AFAIK I merged both branches (master is actually a cherry pick from one-5.10):

one-5.10:
```
commit b19511f4b9208253f9e1f40da165905b9285858f
Author: Kristian Feldsam <feldsam@gmail.com>
Date:   Thu Apr 16 15:37:28 2020 +0200

    B #4148: Added missing showback xsd (#4149)
    
    Signed-off-by: Kristian Feldsam <feldsam@gmail.com>
```

master:
```
commit 3d29bb672fa6407b3a5e0cbb71235a99dc449e9f
Author: Kristian Feldsam <feldsam@gmail.com>
Date:   Thu Apr 16 15:37:28 2020 +0200

    B #4148: Added missing showback xsd (#4149)
    
    Signed-off-by: Kristian Feldsam <feldsam@gmail.com>
    (cherry picked from commit b19511f4b9208253f9e1f40da165905b9285858f)
```

rsmontero:
Maybe, are you referring to a different commit?

feldsam:
yes, this was merged, but after while reverted in this PR #4095 where is a discussion about it and finally, we decided to use `xs:string` in PR #4117

rsmontero:
Merged it to one-5.10:

```
[...]

PR #4095, opened 2020-01-13
B #4076 Keep xs:integer type for acl

xorel:
Although the format is hexadecimal, ONE produces
values where length is not even (without adding
leading zero). This violates xs:hexBinary type.

### DIFF
--- a/share/doc/xsd/acl_pool.xsd
+++ b/share/doc/xsd/acl_pool.xsd
@@ -8,10 +8,10 @@
           <xs:complexType>
             <xs:sequence>
               <xs:element name="ID" type="xs:integer"/>
-              <xs:element name="USER" type="xs:hexBinary"/>
-              <xs:element name="RESOURCE" type="xs:hexBinary"/>
-              <xs:element name="RIGHTS" type="xs:hexBinary"/>
-              <xs:element name="ZONE" type="xs:hexBinary"/>
+              <xs:element name="USER" type="xs:integer"/>
+              <xs:element name="RESOURCE" type="xs:integer"/>
+              <xs:element name="RIGHTS" type="xs:integer"/>
+              <xs:element name="ZONE" type="xs:integer"/>
               <xs:element name="STRING" type="xs:string"/>
             </xs:sequence>
           </xs:complexType>


xorel:
merge to
* master
* one-5.10

feldsam:
but it returns hex, so integer is wrong to use. script fails on fatal error, so you can use string or define own hex type.

feldsam:
btw, I tested pyone with xs:hexBinary and it is working, don't know how other languages.

xorel:
Our test suite which uses xmllint for validation xml fails against xsd was failing due to the not even length of the hex string (perhaps missing leading zero), so I reverted back.

 > script fails on fatal error

You mean your the python script fails with xs:integer?

feldsam:
ok, I understand. Problem is, that pyone fails on error ~"integer excepted, got string", because during build time some "supbind" code is generated from XSD schema...

The simple solution is to use `xs:string` instead of `xs:integer`

xorel:
I am fine with changing to `xs:string`. But still, could you send me a snippet of failing code? I can't reproduce somehow.

feldsam:
Hi, problem is that you don't receive ACL data, because ACL is missing in index.xsd. So I added index XSD and rebuild pyone and got error about string instead excepted integer. With actual package in PIP you get "HISTORY_RECORDS" instead of ACL pool.


```
from pyone import OneServer

one = OneServer('https://api.feldcloud.net/RPC2', session='%s:%s' % ('user', 'password'))

acl_pool = one.acl.info()

print(acl_pool)
```

Try to rebuild locally pyone with included ACL in index.xsd and you hit error.

xorel:
I still can not reproduce, I have fresh pyone built with index.xsd referencing acl_pool and 
```           <xs:element name="ID" type="xs:integer"/>                                                                                                               
              <xs:element name="USER" type="xs:integer"/>
              <xs:element name="RESOURCE" type="xs:integer"/>
              <xs:element name="RIGHTS" type="xs:integer"/>
              <xs:element name="ZONE" type="xs:integer"/>
              <xs:element name="STRING" type="xs:string"/>
```
Then I can easily run your code or iterate ACLs and call `.get_USER()` or similar (trying python2 and python3 as well).

feldsam:
Did you try `one.acl.info()` ? In some situations I can imagine that it can work, because of only integer chars in hex.

What `one.acl.info()` returns you?

xorel:
I get:
<pyone.bindings.ACL_POOLSub object at 0x7f4bee4b77b8>

feldsam:
and what is content?

xorel:
What do you man? When I iterate over ACLs like this
```
for acl in one.acl.info().ACL:
    print("{}".format(acl.get_USER()))
```
I get:
```
200000001
400000000
400000000
200000001
...```

feldsam:
you are lucky, you get only integers in HEX...you need to add some more ACLs which retuns hex with letters too

feldsam:
can I test too? is pyone with acls published on PIP?

xorel:
OK, you are right, I got it. Let me then rollback to xs:string and sorry for the noise.

For the PIP no, we do releases of pyone together with opennebula release.

feldsam:
@xorel Hello Honza, this is still not in the one-5.10 branch, only in master. Why these PyOne issues are not fixed in hotfix releases too?

PR #4101, opened 2020-01-15
B 4087: Fix incorrect disk keys

mouyaq:


src/vmm_mad/remotes/lib/vcenter_driver/virtual_machine.rb
```diff
@@ -1663,7 +1695,7 @@ def sync_disks(option = :nil, execute = true)
 
             spec = RbVmomi::VIM.VirtualMachineConfigSpec(spec_hash)
             @item.ReconfigVM_Task(:spec => spec).wait_for_completion
-
+            ###sync_extraconfig_disk(spec_hash)
```
tinova: can we remove this?

tinova:
also in one-5.10

Issue #4087, opened 2020-01-08
lost opennebula managed disk after migrate changing datastore


mouyaq:
**Description**
After doing a migrate opennebula managed disk can be detached from the instance, and under some conditions the disks could be removed.

**To Reproduce**
Deploy an instance from an imported vcenter template
Create a new image ( an empty datablock for example )
Attach the image to the instance
Power off the instance
Migrate and 
Power on the instance
The attached disk will be renamed and moved to the destination datastore and an error will appear at powering on because the instance will try to find the old disk path.

**Expected behavior**
Disk path should be updated correctly

**Details**
 - Affected Component: [vCenter driver]
 - Hypervisor: [ESXi]
 - Version: [e.g. 5.4.12 or development]

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

mouyaq:
PR: https://github.com/OpenNebula/one/pull/4101

tinova:
This will be properly tested in the context of #4105. Closing for now.

PR #4106, opened 2020-01-17
B #4087: Allows attach IDE disks

mouyaq:
This allows attach IDE disk to VMS when the VM is
into powered off state.
The disk will have DEV_PREFIX="sd" by default, if you want
to have a DEV_PREVIX="hd" you need to change before into the
image template.


tinova:
taken to one-5.10

Issue #1269, opened 2017-11-20
vCenter Resource pool tab in host


OpenNebulaProject:
---

Author Name: **Tino Vázquez** (Tino Vázquez)
Original Redmine Issue: 5240, https://dev.opennebula.org/issues/5240
Original Date: 2017-07-13

---

Also, expose the Resource Pool functionality in the VM Template and VM instantiation dialogs

carloshm91:
[DEV]
* https://github.com/OpenNebula/development/pull/829

[ONE]
* https://github.com/OpenNebula/one/pull/4108

[DOC]
* https://github.com/OpenNebula/docs/pull/913

PR #4108, opened 2020-01-20
F #1269: show resource pool inside cards

carloshm91:
-> master
-> one-5.10


tinova:
only in master, due to the changes in configuration files

Issue #4109, opened 2020-01-20
Sync configuration files in an HA environment


al3xhh:
**Description**
Command to sync OpenNebula's configurations across different federated and High Availability (HA) zones and fix lagging nodes in HA environments.

**Use case**
It's useful to help cloud administrator to sync all the files.

**Interface Changes**
CLI

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
Code: https://github.com/OpenNebula/one/pull/4116
Docs: https://github.com/OpenNebula/docs/pull/902 https://github.com/OpenNebula/docs/pull/903
Tests: https://github.com/OpenNebula/development/pull/820

PR #4116, opened 2020-01-22
F #4109: add onezone serversync command

al3xhh:
Applies to 5.10 also.

**Do not squash commits**

### DIFF
--- a/src/cli/one_helper/onezone_helper.rb
+++ b/src/cli/one_helper/onezone_helper.rb
@@ -14,7 +14,433 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
+require 'fileutils'
+require 'tempfile'
+require 'CommandManager'
+
 require 'one_helper'
+
+# Check differences between files and copy them
+class Replicator
+
+    SSH_OPTIONS = '-o stricthostkeychecking=no -o passwordauthentication=no'
+    ONE_AUTH    = '/var/lib/one/.one/one_auth'
+    FED_ATTRS   = %w[MODE ZONE_ID SERVER_ID MASTER_ONED]
+
+    FILES = [
+        { :name    => 'az_driver.conf',
+          :service => 'opennebula' },
+        { :name    => 'az_driver.default',
+          :service => 'opennebula' },
+        { :name    => 'ec2_driver.conf',
+          :service => 'opennebula' },
+        { :name    => 'ec2_driver.default',
+          :service => 'opennebula' },
+        { :name    => 'econe.conf',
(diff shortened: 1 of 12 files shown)


src/cli/one_helper/onezone_helper.rb
```diff
@@ -14,8 +14,373 @@
...
+    end
+
+    # Service action
+    #
+    # @param service [String] Service to restart
+    # @param action  [String] Action to execute (start, stop, restart)
+    def service_action(service, action = 'restart')
+        if `file /sbin/init`.include? 'systemd'
+            puts "Restarting #{service} via systemd"
+            run_command("systemctl #{action} #{service}")
```
vholer: Better `restart` -> `try-restart` OR you have to check first if the service is running. Also, the case when Apache is running within Apache is not handled (at least with `try-restart` won't start another Sunstone instance).

src/cli/one_helper/onezone_helper.rb
```diff
@@ -14,8 +14,373 @@
...
+    # @param folder  [String] Folder to copy
+    # @param service [String] Service to restart
+    def copy_folder(folder, service)
+        puts "Checking #{folder}"
+
+        output = run_command(
+            "rsync -ai\
+            -e \"ssh #{SSH_OPTIONS} -i #{@oneadmin_identity_file}\" " \
+            "oneadmin@#{@remote_server}:/etc/one/#{folder}/ " \
+            "/etc/one/#{folder}/"
```
vholer: One comment for all the remote calls, I don't think you can rely on the fact that you can connect to oneadmin on another frontend. Also, the accesses are mixed, somewhere the `oneadmin` is enforced, somwhere the current user (e.g., `SSHCommand`) is used. So, ideally it should try the unspecified current user, if it doesn't work, it can try oneadmin, if it doesn't work, it should fail.

src/cli/one_helper/onezone_helper.rb
```diff
@@ -14,8 +14,373 @@
 # limitations under the License.                                             #
 #--------------------------------------------------------------------------- #
 
+require 'augeas'
```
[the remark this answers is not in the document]
vholer: BTW - It would be nice if this is not hard dependency for whole onezone command, but just required for the serversync subcommand. Also, it should fail with some nice error message "ERROR: You need to have Ruby augeas gem installed .... following way ...." if the dependency is not fulfiled.

(Ideally we should also have this new Augeas dependency in opennebula-cli gems, but 99.9% of users don't care about this new serversync feature and require them all to install Augeas devel looks like a big pain to me.

PR #4135, opened 2020-01-29
[GOCA] permissions changes

treywelsh:
update Permissions structure fields types (without breaking changes)


tinova:
also, to one-5.10

Issue #4154, opened 2020-02-04
Memory leak after upgrade to 5.10.1


MedicMomcilo:
**Description**
We ran 5.8 for long period and decided to upgrade.
After upgrade, RAM usage is really high (over 20GiB) with OOM kicking in.

We followed the guide for upgrade and faced no issues during.

**To Reproduce**
Upgrade the cluster.

**Expected behavior**
Similar RAM usage/requirement.

**Details**
 - Affected Component: oned
 - Hypervisor: KVM
 - Version: 5.10.1

**Additional context**
Frontend runs on a dedicated physical host with 32GiB of RAM and 4 cores.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
We've uploaded a patch for a couple of leaks. However they are "minor" and we are checking for additional problems.

We assume that 5.8 behavior was ok, in terms of memory consumption?

MedicMomcilo:
Yes, memory consumption was not an issue with 5.8.
This looks like a major leak as we ran out of 32GiB of RAM + 16GiB of SWAP.

It gets worse during the night where we have heavy ONE activity:
- nightly snapshots for all CEPH-based VMs (creating and removing snapshots)
- image backups for large number of VMs (shutting down, snapshot, booting,...)

All of those come with a ton of side-requests for VM states, disk type information, etc.

rsmontero:
So we've been able to reproduce the issue. It seems that those ton of side-requests generate events at a pace much greater that the current event driver can handle. This, in turns makes oned buffers to increase and eventually OOM takes part in the party.

As those API calls are info (and currently we do not see a clear use case for them) we have disabled events on those API calls. We have uploaded a patch to master and one-5.10 branch. 

Thanks for your feedback and detailed comments. Closing the issue

PS: As a side effect we have fixed those minor leaks :)

MedicMomcilo:
Hi Ruben,

Unfortunately this doesn't solve memory leak on our end.
```
root@one0:~# oned --version
Copyright 2002-2019, OpenNebula Project, OpenNebula Systems        

OpenNebula 5.10.2 (not known) is distributed and licensed for use under the terms of the
[...]

PR #4157, opened 2020-02-05
B #4154: marks info call as unsupported

christian7007:


tinova:
we should use this issue number: https://github.com/OpenNebula/one/issues/4154

christian7007:
Done.

Issue #4164, opened 2020-02-07
detaching disk issue: "error detaching VM Disk: Could not detach sdc from"


pumpkineggs:
**Description**
While detaching disk from VM in some cases an error could be thrown like: "error detaching VM Disk: Could not detach sdc from <domain>". 

**To Reproduce**
1) Create user with name sdaniel
2) Create a VM owned by the user and attach scsi disk (sda).
3) Try to detach the disk on running VM.
4) Get an error.

**Expected behavior**
Disk detached, no error thrown.

**Details**
 - Affected Component: vmm_mad 
 - Hypervisor: KVM
 - Version: 5.10.x

**Additional context**
We noticed that "detach_disk" script (vmm_mad/remotes/kvm/detach_disk) should use a more efficient way to check whether a disk detached or not.
While detaching disk from VM in some cases an error could be thrown like: "error detaching VM Disk: Could not detach sdc from <domain>". 
But in fact, the disk has been detached from the VM and it can be confirmed via "virsh dumpxml" command.
There is a piece of code in the script:
```
virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | grep $TARGET > /dev/null 2>&1

if [ $? -eq 0 ] ; then
    error_message "Could not detach $TARGET from $DOMAIN"
    exit -1
fi
```
Here we have ```grep $TARGET``` part where $TARGET turns into a device name (i.e. vda, sda, sdc, etc...).
Let's say we are trying to detach a scsi disk "sda", so the script will run a check with ```grep sda```.
And an interesting part is if the owner of the VM is a user with a name: sdaniel, which is reflected in the domain xml metadata section, grep will return some data and ```if [ $? -eq 0 ] ;``` piece will return an error and script will exit with rc=-1 in spite the disk has been detached from the domain. 

Proposed changes for the script:
```virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | grep dev=\'$TARGET\' > /dev/null 2>&1```
Just be more specific about what to grep and extend the regexp to ```dev=\'$TARGET\'```.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
merged

PR #4168, opened 2020-02-07
B #4164: Use xmllint to test disk target device

atodorov-storpool:

### DIFF
--- a/src/vmm_mad/remotes/kvm/detach_disk
+++ b/src/vmm_mad/remotes/kvm/detach_disk
@@ -29,9 +29,10 @@ DETACH_PARAMS="--domain $DOMAIN --target $TARGET"
 exec_and_log "virsh --connect $LIBVIRT_URI detach-disk $DETACH_PARAMS" \
     "Could not detach $TARGET from $DOMAIN"
 
-virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | grep $TARGET > /dev/null 2>&1
+virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | \
+    xmllint -xpath "//disk/target[@dev='$TARGET']" - >/dev/null 2>&1
 
-if [ $? -eq 0 ] ; then
+if [ $? -ne 0 ] ; then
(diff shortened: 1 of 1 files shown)


src/vmm_mad/remotes/kvm/detach_disk
```diff
@@ -29,9 +29,10 @@ DETACH_PARAMS="--domain $DOMAIN --target $TARGET"
 exec_and_log "virsh --connect $LIBVIRT_URI detach-disk $DETACH_PARAMS" \
     "Could not detach $TARGET from $DOMAIN"
 
-virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | grep $TARGET > /dev/null 2>&1
+virsh --connect $LIBVIRT_URI dumpxml $DOMAIN | \
+    xmllint -xpath "//disk/target[@dev='$TARGET']" >/dev/null 2>&1
```
kvaps: This will adds `xmllint` to dependencies, wouldn't be better to use `/var/tmp/one/datastore/xpath.rb` instead?

atodorov-storpool: The following reasons:
* `xmllint` is part of the `libxml2` on which several `libvirt` sub-packages depends
* the exit code of `xpath.rb` could not be used. I hate adding unnecessary piping/processing just to get an exit status.

atodorov-storpool: ... and `augeas` ;-)
So it is already available due to required dependencies.

xorel: True, so I also prefer using `xmllint`. One small thing, there seems to be missing `-` (hyphen sign) as a last xmllint parameter to even process the stdin.

atodorov-storpool: hm. stupid typo :)
how do you prefer - to push second commit or edit and force-push?

xorel: Doesn't really matter, you can force-push or add another commit, which will be squashed when merging.

Issue #3859, opened 2019-10-18
New Monitoring System


paczerny:
**Description**
Monitoring system has limited scalability. For huge installations monitoring can slow down performance of the whole system.

Basic idea is to separate monitoring from oned to new monitoring daemon. Store monitoring data in separate (time series) DB. Reduce number of DB accesses.

**Use case**
Increase number of Host and VMs handled by oned

**Interface Changes**
xml-rpc API one.host.info and one.vm.info should have parameter to include monitor information.

**Additional Context**
More details [monitor design](https://docs.google.com/document/d/1cIw3lq3wwlzqlbGgxH2_1GakzIVk6uBSg9VcUiTtbaQ/edit#heading=h.q0xwvg9ykqqm) 

Also consider additional status mappings for every hipervisor in order to handle hipervisor-level status changes, for example, a VM that paused due to hipervisor node running out of space. This also would require new actions on the vmm drivers, like pausing and unpausing.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

kvaps:
> Store monitoring data in separate (time series) DB. Reduce number of DB accesses.
Consider also storing VM logs in tsdb

PR #4191, opened 2020-02-14
More accurate datastore/ space monitoring

mzealey:
When monitoring a host's overall datastore/ disk space usage, look at
the filesystems for each datastore (which may be mounted) and sum the
total amounts of free/used/total space for all different mounts under
that directory. This should allow more accurate vm scheduling based on
overall disk space usage on a particular host.

### DIFF
--- a/src/im_mad/remotes/common.d/monitor_ds.sh
+++ b/src/im_mad/remotes/common.d/monitor_ds.sh
@@ -25,19 +25,13 @@ REMOTES_DIR="${DRIVER_PATH}/../.."
 
 mkdir -p "$DATASTORE_LOCATION"
 
-USED_MB=$(df -B1M -P $DATASTORE_LOCATION 2>/dev/null | tail -n 1 | awk '{print $3}')
-TOTAL_MB=$(df -B1M -P $DATASTORE_LOCATION 2>/dev/null | tail -n 1 | awk '{print $2}')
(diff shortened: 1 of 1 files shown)


rsmontero:
INTERNAL NOTE: merge after new monitoring (#3859) branch is merged

mzealey:
Is there a reason this small improvement couldn't be merged prior to that ie next 5.10 point release?

rsmontero:
Totally, didn't take a look at the patch itself....

xorel:
This looks reasonable, however I suggest to keep the monitor script bash-only.

Can you take a look at https://github.com/xorel/one/commit/77080463cafbc0e845e1124ded701d05a3b80062 ?

mzealey:
This looks like a reasonable approach however `stat $dir --printf=%m` only prints the mount root or something like that of the filesystem which is different if I have several bind-mount points pointing at the same filesystem.

Replacing the stat commands with something like `stat -f $dir --printf=%i` works as expected.

xorel:
Right, the `%m` doesn't work well for bind-mounts. The %i seems to work better, but for the sake of simplicity, I modified to read the source from `df`.  Can you check if https://github.com/xorel/one/commit/cbef7749945a911a71d4e103bf14a2b28da02904 works for you?

xorel:
I made some further changes in monitor_ds, like avoiding multiple running of `df` or unify var casing.
If you don't mind I will close this PR and proceed with https://github.com/OpenNebula/one/pull/4243

mzealey:
Sounds good thanks!

Issue #1542, opened 2017-11-28
vCenter hot disk resize


tinova:
# Enhancement Request
## Description

vCenter allows for hot disk resize, OpenNebula should as well

## Use case

Want to grow my disk while VM is running (probably makes sense for non OS disks)

## Interface Changes

Options in Sunstone and CLI

xBytez:
+1 This is really something critical we need in order to start using one in production.

tinova:
See what support comes from OpenNebula core

See what support do we have in current vCenter drivers

xBytez:
I hacked around in sunstone and removed the frontend if checks that disables live disk resizing for VMware. Nebula resizes the disk live just fine. 

The VM just gets stuck in DISK_RSZ status which you can fix with `onevm recover id --success`

tinova:
Thanks a lot for the feedback! Looks like we are close ;)

carloshm91:
[ONE]
* https://github.com/OpenNebula/one/pull/4199

[DOC]
* https://github.com/OpenNebula/docs/pull/945

[TEST]
* https://github.com/OpenNebula/development/pull/847

PR #4199, opened 2020-02-17
F #1542: vCenter hot disk resize

carloshm91:
-> máster

Signed-off-by: Carlos Herrera <cherrera@opennebula.systems>


tinova:
also to one-5.10

PR #4202, opened 2020-02-18
M #~: Sanitize the code

sbbarragan:
Applies to 5.10 also


tinova:
looks like this is not needed in one-5.10

Issue #4061, opened 2019-12-20
Sunstone: Embed fonts local (not from Google Fonts)


vidister:
**Description**
At the moment the Font "Lato" is fetched from Google Fonts. This makes things slower and collides with sane Content Security Policies.
Please evaluate if bundling the font with the software (as it's already happening with OpenSans at the moment) is an option.

I volunteer for a Pull Request.

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

jloboescalona2:
@vidister  We would be very grateful for your contribution if you do the PR

tinova:
thanks a lot for the contribution and apologies for the huge delay in addressing this. we will make sure this is the default in the new sunstone iteration #5422

PR #4246, opened 2020-02-21
#4061 Sunstone: Embed fonts local

vidister:
At the moment the Font "Lato" is fetched from Google Fonts. This makes things slower and collides with sane Content Security Policies.

This Pull Request addresses that issue by bundling the font with the software.

Lato is licensed under the SIL Open Font License which allows bundling it with software.


vidister:
@jloboescalona2 I do, but I only used woff2 since the font was embedded with woff2 only originally. Should I add the other formats too?
When doing that I also can also add the other formats for the OpenSans font which is only embedded in woff at the moment.

jloboescalona2:
@vidister, Yes, I know that there is only the woff, but it would be interesting to have it all, or those that can be seen in more browsers, I also take the opportunity to thank you for your contribution to the project

sbbarragan:
Closing as this will be recreated from scratch in https://github.com/OpenNebula/one/issues/5422

Issue #4329, opened 2020-03-10
onedb fsck fails with "utf-8" database encoding


vholer:
**Description**

Related to this translation table:
https://github.com/OpenNebula/one/blob/master/src/onedb/onedb_backend.rb#L353

If database is in **utf8** encoding, the `onedb fsck` fails as it doesn't match exact encoding,  but "utf8 is an alias for utf8mb3; the character limit is implicit, rather than explicit in the name."

```
> show create database opennebula;
+------------+---------------------------------------------------------------------+
| Database   | Create Database                                                     |
+------------+---------------------------------------------------------------------+
| opennebula | CREATE DATABASE `opennebula` /*!40100 DEFAULT CHARACTER SET utf8 */ |
+------------+---------------------------------------------------------------------+
```

Also, catch-all encoding **NONE** is expected to fail. Probably it should fail more nicely in advance, or maybe always default to ASCII (???).

```
$ onedb fsck -v -f  -u 'oneadmin'  -p 'XXXXX'  -d 'opennebula'
Version read:
Shared tables 5.10.0 : OpenNebula 5.10.1 (45f6d6df) daemon bootstrap
Local tables  5.10.0 : OpenNebula 5.10.1 (45f6d6df) daemon bootstrap

MySQL dump stored in /var/lib/one/mysql_localhost_opennebula_2020-3-10_8:50:10.sql
Use 'onedb restore' or restore the DB using the mysql command:
mysql -u user -h server -P port db_name < backup_file

  > Running fsck


unknown encoding name - NONE
/usr/share/one/gems-dist/gems/nokogiri-1.10.5/lib/nokogiri/xml/node.rb:697:in `find'
/usr/share/one/gems-dist/gems/nokogiri-1.10.5/lib/nokogiri/xml/node.rb:697:in `serialize'
/usr/share/one/gems-dist/gems/nokogiri-1.10.5/lib/nokogiri/xml/node.rb:722:in `to_xml'
/usr/share/one/gems-dist/gems/nokogiri-1.10.5/lib/nokogiri/xml/node.rb:583:in `to_s'
```

Because,

```
$ irb
irb(main):001:0> Encoding.find('NONE')
Traceback (most recent call last):
        3: from /usr/bin/irb:11:in `<main>'
        2: from (irb):1
        1: from (irb):1:in `find'
ArgumentError (unknown encoding name - NONE)
```

**Details**
 - Affected Component: onedb
 - Version: 5.10.3

## Progress Status
[...]

PR #4330, opened 2020-03-10
B #4329: use default encoding if not found

al3xhh:
Applies to 5.10 too


kvaps:
please ship it to 5.10.4

al3xhh:
Hello @kvaps 

We have also merged it into 5.10 branch, so it will be available in the future 5.10.4 release.

PR #4332, opened 2020-03-10
[feature] implement crash detection for qemu vms

philfry:
Hi all,

right now afaik OpenNebula is unable to tell whether a qemu-VM has crashed.
This PR adds the ability to detect VM crashes/panics and to define what action should be taken.

If set, an isa-panic-device is added to the vm as well as the on_crash-action itself:
```xml
<on_crash>{destroy|restart|preserve|coredump-destroy|coredump-restart}</on_crash>
<devices>
  <panic model='isa'/>
</devices>
```


philfry:
Since DCO is complaining: What's the correct way to handle long email addresses that cause the signed-off-by-line to exceed 78 chars? Is a line break okay?

Issue #4334, opened 2020-03-10
onedb fsck count reserverd networks wrong


kvaps:
**Description**

I've already reported that in https://github.com/OpenNebula/one/issues/1710#issuecomment-597268284:

```
Group 3133 quotas: VNet 23	LEASES_USED has 6 	is	0
Group 3133 quotas: VNet 111111	LEASES_USED has 0 	is	6
```

I have no any network with id `111111`

But Group 3133 owns network `5432` with 6 address ranges:

```xml
<VNET>
  <ID>5432</ID>
  <UID>3334</UID>
  <GID>3133</GID>
  <UNAME>ov3211</UNAME>
  <GNAME>ov3211</GNAME>
  <NAME>ov3211</NAME>
  <PERMISSIONS>
    <OWNER_U>1</OWNER_U>
    <OWNER_M>1</OWNER_M>
    <OWNER_A>0</OWNER_A>
    <GROUP_U>0</GROUP_U>
    <GROUP_M>0</GROUP_M>
    <GROUP_A>0</GROUP_A>
    <OTHER_U>0</OTHER_U>
    <OTHER_M>0</OTHER_M>
    <OTHER_A>0</OTHER_A>
  </PERMISSIONS>
  <CLUSTERS>
    <ID>100</ID>
    <ID>101</ID>
    <ID>102</ID>
    <ID>103</ID>
    <ID>104</ID>
    <ID>105</ID>
  </CLUSTERS>
  <BRIDGE><![CDATA[vmbr0]]></BRIDGE>
  <BRIDGE_TYPE><![CDATA[linux]]></BRIDGE_TYPE>
  <PARENT_NETWORK_ID>23</PARENT_NETWORK_ID>
  <VN_MAD><![CDATA[802.1Q]]></VN_MAD>
  <PHYDEV><![CDATA[bond0]]></PHYDEV>
  <VLAN_ID><![CDATA[100]]></VLAN_ID>
  <OUTER_VLAN_ID/>
  <VLAN_ID_AUTOMATIC>0</VLAN_ID_AUTOMATIC>
  <OUTER_VLAN_ID_AUTOMATIC>0</OUTER_VLAN_ID_AUTOMATIC>
  <USED_LEASES>5</USED_LEASES>
  <VROUTERS/>
  <TEMPLATE>
    <BRIDGE><![CDATA[vmbr0]]></BRIDGE>
    <BRIDGE_TYPE><![CDATA[linux]]></BRIDGE_TYPE>
[...]

PR #4335, opened 2020-03-10
B #4334: fix reserved networks count

kvaps:
Use `PARENT_NETWORK_ID` instead of `AR/PARENT_NETWORK_AR_ID` to find parent network id

fixes https://github.com/OpenNebula/one/issues/4334


kvaps:
@al3xhh please review that

Issue #4214, opened 2020-02-19
Provision support of other object types


vholer:
**Description**
Provision can create as part of provision cluster, hosts, datastores and vnets (http://docs.opennebula.org/5.10/advanced_components/ddc/reference/provision/overview.html#provision-objects). We might need to extend the support for following more objects:

- templates
   - vm
   - vnet
- images
- marketapp
- oneflow template

All entities should be able to support also configuring:
- owner / group
- permissions mode

Notes:
- images and marketapps should support 2 operation modes
    - **asynchronous** -  just create the objects and continue (exit)
    - **synchronous** - create objects and wait until they are successfully imported
        - in case of failure, retry should delete the failed image and retry create again
- images/marketapp should be able to reference the newly created datastores, probably using ERB
- IMPORTANT: we now merge defaults+overrides for each supported objects (which makes perfect sense for infrastructure objects). I'm not sure if that makes sense to follow that with vnet template, but definitely NOT with end user objects (VM templates, images/marketapps).

**Use case**
New cluster with all necessary objects ready to start VM.

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
PRs to merge in master:

- code: https://github.com/OpenNebula/one/pull/4374
- docs: https://github.com/OpenNebula/docs/pull/1001
- tests: https://github.com/OpenNebula/development/pull/891

PR #4374, opened 2020-03-18
F #4214: provision support of other objects types

al3xhh:


src/oneprovision/lib/provision.rb
```diff
@@ -163,7 +184,8 @@ def status
         # @param config  [String]  Path to the configuration file
         # @param cleanup [Boolean] True to delete running VMs and images
         # @param timeout [Integer] Timeout for deleting running VMs
-        def create(config, cleanup, timeout)
+        # @param skip    [Symbol]  Skip provision, config or none phases
+        def create(config, cleanup, timeout, skip)
```
vholer: Please update this method so there is just single create action for each entity and method **ends at its end**, not in the middle by 2 extra checks/creates and returns.

src/oneprovision/lib/provision.rb
```diff
@@ -186,35 +208,33 @@ def create(config, cleanup, timeout)
...
-
-                    return 0
-                end
-
                 create_hosts(cfg, cid)
 
-                # ask user to be patient, mandatory for now
-                STDERR.puts 'WARNING: This operation can ' \
-                    'take tens of minutes. Please be patient.'
+                create_virtual_resources(cfg)
```
vholer: Virtual resources should be created at the end after hosts are created and configured.

++ create steps order in the docs must be updated

Issue #2506, opened 2018-10-15
Improve filtering matching for elements with multiple values on a column


sergiojvg:
**Description**
Currently, when a filter is applied to an element with multiple values on the same column, it will only return the element if all the values where specified on the filter. It should return the element by specifying only one value.

**Use case**
If vnet 1 is associated to clusters 100 and 101 it will currently only show up if you filter for "100,101". I would expect it to be visible if filtering for 100, or 101 or "100,101".

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
Also, check these kind of issues:

```
# onemarket list --csv --no-header --list NAME --filter ID!=0
Traceback (most recent call last):
	12: from /usr/bin/onemarket:39:in `<main>'
	11: from /usr/bin/onemarket:39:in `new'
	10: from /usr/lib/one/ruby/cli/command_parser.rb:84:in `initialize'
	 9: from /usr/lib/one/ruby/cli/command_parser.rb:470:in `run'
	 8: from /usr/bin/onemarket:147:in `block (2 levels) in <main>'
	 7: from /usr/lib/one/ruby/cli/one_helper.rb:810:in `list_pool'
	 6: from /usr/lib/one/ruby/cli/one_helper.rb:636:in `list_pool_table'
	 5: from /usr/lib/one/ruby/cli/cli_helper.rb:368:in `show'
	 4: from /usr/lib/one/ruby/cli/cli_helper.rb:482:in `print_table'
	 3: from /usr/lib/one/ruby/cli/cli_helper.rb:568:in `data_array'
	 2: from /usr/lib/one/ruby/cli/cli_helper.rb:808:in `filter_data!'
	 1: from /usr/lib/one/ruby/cli/cli_helper.rb:808:in `map'
/usr/lib/one/ruby/cli/cli_helper.rb:822:in `block in filter_data!': undefined local variable or method `left' for #<CLIHelper::ShowTable:0x00005573e61dcf40> (NameError)
```

al3xhh:
PRs to merge in 5.10 and master:

- code: https://github.com/OpenNebula/one/pull/4376
- docs: https://github.com/OpenNebula/docs/pull/1002

PR #4376, opened 2020-03-18
F #2506: improve CLI filtering matching

al3xhh:
New operator has been added for filtering in the CLI.

The new operator '~' will match sub keys, not exact, e.g:

If NAME is 'testing' with --filter NAME=tes wont't match but NAME~tes yes


al3xhh:
Applies to 5.10 also

Issue #1193, opened 2017-11-20
Differentiate created templates on vCenter


OpenNebulaProject:
---

Author Name: **Adrian Menendez** (Adrian Menendez)
Original Redmine Issue: 5053, https://dev.opennebula.org/issues/5053
Original Date: 2017-03-09

---

When create a template making a vm persistent, add an attribute to know if has alredy been created on vCenter (vm has been deleted) or is still only in OpenNebula. Another option is to add an attribute while is waiting and delete it when it's created, so all of the created templates have the same attributes and the only different ones are those who has been created making a vm as persistent.

OpenNebulaProject:
---

Original Redmine Comment
Author Name: **Jaime Melis** (@jmelis)
Original Date: 2017-06-08T09:42:05Z

---

Can you please elaborate? We are having difficulties understanding this issue.

OpenNebulaProject:
---

Original Redmine Comment
Author Name: **Adrian Menendez** (Adrian Menendez)
Original Date: 2017-06-09T10:21:46Z

---

When a vm is deployed as persistent, OpenNebula creates a template too. Meanwhile the vm is up and running, the template is only on OpenNebula, just a registry, not a real vCenter template. It would be interesting to be able to differentiate which of them really exists on vCenter.

tinova:
Check if VM Template in OpenNebula is locked until the vCenter VM is converted to VM Template. If not, lock it.

With this functionality, locked VM Templates are the ones not present in vCenter.

carloshm91:
[ONE]
* https://github.com/OpenNebula/one/pull/4394

[DOC]
* https://github.com/OpenNebula/docs/pull/1064

PR #4394, opened 2020-03-22
F #1193: differentiate created templates on vCenter

carloshm91:
-> master
-> one-5.10

Signed-off-by: Carlos Herrera <cherrera@opennebula.systems>


tinova:
it's a delicate change, let's keep it for 5.12 only

Issue #4406, opened 2020-03-24
Service scale UP VMs don't get the current user and group


OpenNebulaSupport:
**Description**
If a chmod is performed in a service, all VM's get the new group. But if any VMs are added after that by scaling they do not get this new user group.

**To Reproduce**
Chmod a running service and then scale it up.

**Expected behavior**
New VMs should get the current user and group.

**Details**
 - Affected Component: Flow server
 - Version: 5.10 and development.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
PRs to merge in 5.10:

- code: https://github.com/OpenNebula/one/pull/4407
- docs: https://github.com/OpenNebula/docs/pull/1009

PR to merge in master:

- code: https://github.com/OpenNebula/one/pull/4408

al3xhh:
PRs to merge in 5.10:

- code: https://github.com/OpenNebula/one/pull/4423
- docs: https://github.com/OpenNebula/docs/pull/1009

PR to merge in master:

- code: https://github.com/OpenNebula/one/pull/4424

PR #4408, opened 2020-03-24
B #4406: change uid and gid when scaling

al3xhh:


al3xhh:
This PR applies only to **master**

Issue #4089, opened 2020-01-09
Add microVM support via Firecracker


christian7007:
**Description**
Create new Firecracker driver for supporting microVMs.

**Use case**
Allow microVMs deployment as an alternative for VMs and Containers.

**Interface Changes**
This new "hypervisor" should be added to Sunstone interface.

**Additional Context**
Firecracker repo: https://github.com/firecracker-microvm/firecracker
Firecracker docs: https://github.com/firecracker-microvm/firecracker/tree/master/docs

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

tinova:
reopening for Sunstone interface changes

PR #4412, opened 2020-03-24
F #4089: Firecracker node packages

christian7007:


share/sudoers/sudoers.rb
```diff
@@ -52,10 +52,9 @@ def initialize(lib_location)
                 'ip'
             ],
             :MARKET => %W[#{lib_location}/sh/create_container_image.sh],
-            :FC     => %w[/usr/bin/jailer
-                          mount
-                          /var/tmp/one/vmm/firecracker/map_context.sh
-                          /var/tmp/one/vmm/firecracker/clean.sh]
+            :FIRECRACKER => %w[/usr/bin/jailer
+                               mount
+                               one-clean-firecracker-domain]
```
vholer: Have absolute path for cleanup script.

share/pkgs/sudoers/opennebula-node-firecracker
```diff
@@ -0,0 +1 @@
+oneadmin ALL=(ALL:ALL) NOPASSWD: ONE_FIRECRACKER
```
vholer: Is this package going to be installed alongside with other package (e.g., node-kvm?).
If not, probably you'll need some networking command lists (e.g., ONE_NET, ONE_OVS).

christian7007: For the moment the idea is to install it along with KVM packages as it also required KVM.

Issue #3600, opened 2019-08-19
PostgreSQL Support


vholer:
**Description**

OpenNebula uses MySQL and SQLite as DB backends to persist state. State is managed through basic SQL queries. The goal is to implement a new backend for the OpenNebula based on PostgreSQL. This is a tracking issue for the student's bachelor project.

**Milestones**

- Implementation SQL backend based on PostgreSQL
- Update user-land tools that interface the DB directly
- Simple document on how to configure/use PostgreSQL with OpenNebula.

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [x] Documentation 
- [x] Release notes - resolved issues, compatibility, known issues
- [x] Code committed to upstream release/hotfix branches
- [x] Documentation committed to upstream release/hotfix branches

vholer:
New database backend support is merged and documented. Closing.

PR #4454, opened 2020-03-31
F #3600: Initial PostgreSQL Support

IgorSivy:
This pull request introduces PostgreSQL support. It adds PostgreSQL backend implementation in oned and onedb. PostgreSQL is not fully compatible with MySQL or SQLite so a few changes had to be made to the syntax of SQL queries used in the code. Some changes are done through preprocessing of SQL queries in the PostgreSQL backend, through extending the SqlDB interface or changed directly in the code. These include:

-  incompatible data types: MEDIUMTEXT/LONGTEXT -> TEXT, BIGINT UNSIGNED -> NUMERIC(20)

-  REPLACE INTO query -> INSERT INTO query with ON CONFLICT clause

-  LIMIT m,n syntax -> LIMIT n OFFSET m

-  INNER JOIN ... WHERE -> INNER JOIN ... ON

-  1,0 as boolean values must be quoted -> '1', '0'

-  MySQL quotes \`...\` -> "..."

Possible future improvements include:

-  "user" column in acl had to be quoted as it is PostgreSQL reserved keyword, which now breaks the MySQL backend because of different quoting styles. It would be best to rename the column.

-  not all incompatible data types are preprocessed, only those that are used in SQL queries used in the code

-  full-text search is currently not supported

-  mysql2postgresql migrator


vholer:
PRs to merge alongside:
- packages https://github.com/OpenNebula/packages/pull/132
- docs https://github.com/OpenNebula/docs/pull/1029

rsmontero:
This is now merged. The merge include a rename of the acl column from user to userset. Also some simplification where made to the LIMIT clause by pushing the offset and number of rows to the DB implementation classes. 

THANKS for this contribution :)

Issue #4128, opened 2020-01-28
Allow multiple hooks for same trigger event


OpenNebulaSupport:
**Description**
It would be useful to be able to have multiple hooks for the same event in order to split hook scripts functionality.

**Additional Context**
Currently the HEM assumes that there is just one hook for any event.

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

PR #4457, opened 2020-03-31
B #4128: Allow multiple hooks for same trigger event

divanikus:


src/hem/onehem-server.rb
```diff
@@ -312,7 +314,10 @@ def load_hook(hook_id)
     def delete(hook_id)
         hook = @hooks_id[hook_id]
 
-        @hooks[hook.type].delete(hook.key)
+        if @hooks[hook.type].key?(hook.key)
+            @hooks[hook.type][hook.key].delete(hook_id)
+            @hooks[hook.type].delete(hook.key) if @hooks[hook.type][hook.key].empty?
+        end
         @filters.delete(hook_id)
```
christian7007, line 321: You need to remove the hook filter only when `@hooks[hook.type][hook.key].empty?`. If you remove it always you'll stop receiving the event when deleting just one of the hooks.

divanikus: I don't seem to understand you. `@hooks[hook.type][hook.key]` is an array of hook_ids, so the `hook.key` key of the `@hooks[hook.type]` hash is removed only if no more hook_ids left. While `@filters` has just individual hook_ids.

christian7007: OK, I see it. You're right sorry. 

In that case, we are subscribing to the same event multiple times, I think that's not a problem as probably zeromq will just see that the filter exists and don't do anything. What I think that can be a problem is the call to unsubscribe here: https://github.com/OpenNebula/one/blob/e316f5b518abff028fc175619e3a3d6f6d11fd45/src/hem/onehem-server.rb#L488

I'm not sure, but probably when we unsubscribe from one filter we will lost the event for the remaining hooks. Have you check this?

divanikus: I guess you are right, that could be a problem on hook updates, didn't researched that scenario. So, ti mitigate this, we should subscribe only to unique events and also unsubscribe only if no subsciber left. I'll check that I can do about that.

christian7007: I've done some test and its seems that zeromq does take into account the number of times you've subscribe to each filter, so don't worry about this.

I'll prepare some more test and we will merge it.

Thank you very much for your contribution!

christian7007: BTW, I used this simple script for testing it: https://gist.github.com/christian7007/fcd58acfd8e8782657ec7bad03ef9c09

Issue #4503, opened 2020-04-08
seg-fault  while executing xpath expression for non existing object in  the oned or mm_sched


Abdullahaml1:
**Description**
segmentation fault happened when executing xpath expression for a non existing object in the pool. for example `/VM_POOL/VM[ID=555]/UID` if the vm id `555` does not exist the code will stuck because of seg-fault  

**To Reproduce**
for example: 
`object.xpaths(result, "/VM_POOL/VM[ID=555]/UID")`, and  vm id `555` does not exist

**Expected behavior**
If the id or the xml node does not exist it will return and the vector will be empty.

**Details**
 - Affected Component: oned, sched
 - Hypervisor: KVM
 - Version: 5.10.0

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
merged!

THANKS FOR THE CONTRIBUTION!

Abdullahaml1:
YOU ARE WELCOME!

PR #4505, opened 2020-04-08
B #4503: checking nodestval before dereferencing

Abdullahaml1:


rsmontero:
Thanks, we do not have an example that triggers this new condition, but its  a safe consistency check.

Issue #4302, opened 2020-03-04
Change vcenter monitoring


mouyaq:
**Description**
Improve and adapt vcenter moniroting to the new monitoriing system

**Use case**
How are you going to use this new feature? Why do you need it?

**Interface Changes**
Describe any changed you'd like to see to current interfaces including Sunstone, CLI or/and API

**Additional Context**
Add any other context or screenshots about the feature request here. Or any other alternative you have considered to addressed this new feature.

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

tinova:
Pending issues

- [ ] VM snapshots
- [ ] VM nic hotplug
- [ ] VNC VCENTER_ESX_HOST (add to vms/status)
- [ ] db lock
- [ ] Error deploying virtual machine: vcenter: NSX_STATUS is not OK
- [ ] VM migration
- [ ] sunstone error while executing tests (pure virtual method called, terminate called without an active exception)

PR #4708, opened 2020-05-12
F #4302: Change wild detecion method and fix resource pool names

mouyaq:


mouyaq:
Re-dded disk info to vm monitoring

PR #4760, opened 2020-05-19
F #3859: Monitor daemon HA

paczerny:


share/etc/oned.conf
```diff
@@ -185,13 +185,13 @@ RAFT = [
 # Executed when a server transits from follower->leader
 # RAFT_LEADER_HOOK = [
 #     COMMAND = "raft/vip.sh",
-#     ARGUMENTS = "leader <interface> <ip_cidr>"
+#     ARGUMENTS = "leader interface:ip_cidr [interface:ip_cidr ...]"
```
vholer, line 188: This is unfortunate for upgrades (also might not handle properly IPv6 addresses with `:` inside?)

Can we better consider just arguments extension without introducing breaking change? E.g.,
`leader interface ip_cidr interface2 ip_cidr2 interface3 ip_cidr3`

paczerny:
Force pushed new commit:
* Address Vlasta's comment, vip.sh arguments are "leader/follower interface ip [interface ip ...]
* squash all commits into one

src/im/InformationManager.cc
```diff
@@ -187,6 +172,31 @@ void InformationManager::delete_host(int hid)
...
+
+    if (!imd)
+    {
+        return;
+    }
+
+    Message<OpenNebulaMessages> msg;
+
+    msg.type(OpenNebulaMessages::RAFT_STATUS);
+    msg.payload(to_string(leader));
```
rsmontero, line 192: Better use pre-defined constants

RAFT_STATUS <LEADER|FOLLOWER>

src/monitor/src/monitor/OneMonitorDriver.cc
```diff
@@ -153,3 +156,16 @@ void OneMonitorDriver::_stop_monitor(message_t msg)
...
+/* -------------------------------------------------------------------------- */
+
+void OneMonitorDriver::_raft_status(message_t msg)
+{
+    istringstream iss(msg->payload());
+    bool is_leader;
+
+    iss >> is_leader;
+
+    hm->raft_status(is_leader);
```
rsmontero, line 170: Update with oned string LEADER/FOLLOWER

src/im/InformationManager.cc
```diff
@@ -187,6 +172,31 @@ void InformationManager::delete_host(int hid)
 /* -------------------------------------------------------------------------- */
 /* -------------------------------------------------------------------------- */
 
+void InformationManager::raft_status(bool leader)
+{
+    if (leader)
+    {
+        send_host_pool();
```
rsmontero, line 179: send_host_pool is only used here, better put the code here and reuse the get_dirver function call and error handling

src/monitor/src/monitor/HostMonitorManager.cc
```diff
@@ -220,6 +221,14 @@ void HostMonitorManager::stop_host_monitor(int oid)
 /* -------------------------------------------------------------------------- */
 /* -------------------------------------------------------------------------- */
 
+void HostMonitorManager::raft_status(bool leader)
```
rsmontero, line 224: better call this function is_leader and update class attribute _is_leader to follow convention

paczerny: This one conflicts with previous comment to call raft_status with string.
Done everything except this one. Waiting for Jenkins to test in microenv

Issue #3354, opened 2019-05-18
VN_MAD=fw IP spoofing block NIC Aliasses


atodorov-storpool:
**Description**
Added IPs as NIC aliases are blocked by the firewall on the host

**To Reproduce**
Add a NIC alias to VM interface with `VN_MAD=fw` and try to ping the address from internet.

**Expected behavior**
The alias IP should be reachable.

**Details**
 - Affected Component: [Network]
 - Hypervisor: [KVM]
 - Version: [5.8.x]

**Additional context**
The alias IPs are not included in the ipset on the host. Manually adding them to the ipset makes them reachable.
The alias IPs are missing from the libvirt filter too. I mean the _clean-trafic_ rule in the domain XML - there is param entry for the primary ip only.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

atodorov-storpool:
Looks like it is related related to #3079 
But the IP spoofing filter

atodorov-storpool:
The patch for MAC spoofing filter for #3079 is not so trivial, though. At least on first read. So the patch is for this issue only...

atodorov-storpool:
The above patch resolve the case of VM resume.
Attaching nic is not resolved, trying to handle it with VM_HOOK on HOTPLUG_NIC but it is hard due to #3357 

On a side note - the VM's Template passed to the VM_HOOK on HOTPLUG_NIC event contain flag ATTACH=YES when detaching a nic...

atodorov-storpool:
And the [net_fw_hook](https://github.com/OpenNebula/addon-storpool/blob/master/hooks/net_fw_hook) that handle the hotplug add/remove of a NIC alias

rsmontero:
After discussing this with the team, it is probably a better idea to have pre_alias, post_alias, clean_alias scripts to configure any thing related to the alias interface in the network.

atodorov-storpool:
Working on this (also addressing #869)...

But there is no nokogiri on the nodes. So I am in doubt to install _opennebula-ruby_ on the nodes or to install and use the OS package for _ruby-nokogiri_?

atodorov-storpool:
[...]

Issue #3079, opened 2019-03-14
802.1q MAC Spoofing blocks Network Aliases


TinoatTSI:
We are running the latest version 5.8.0 of OpenNebula und Ubuntu 18.04. If you add an alias (Attach nic) to an existing virtual NIC, the communication is dropped, if the not the MAC of the inital interface is used. 
If you e.g. want to run a virtual machine on the VM or something using macvlan, the VM is unable to communication with the network, since the MAC spoofing protection setup by OpenNebula on the host interface will drop the packages.

```
Chain one-21-0-o (1 references)
num  target     prot opt source               destination
1    DROP       all  --  0.0.0.0/0            0.0.0.0/0            MAC ! 03:11:0A:0C:00:04
2    ACCEPT     all  --  0.0.0.0/0            0.0.0.0/0            state RELATED,ESTABLISHED
3    RETURN     all  --  0.0.0.0/0            0.0.0.0/0
4    DROP       all  --  0.0.0.0/0            0.0.0.0/0
>>>>
```

Manually removing rule 1 will work, but is more a dirty quick fix that does not last.

Can you please add additonal MACs also to the filter?

rsmontero:
Totally! Thanks for the feedback. 

As you said the driver needs to be updated to respond to this filters for the ALIAS. In the meanwhile you would need to disable MAC spoofing

rsmontero:
Sorry, I was reading again the issue. We are aware of the IP spoofing (i.e. some of the functionalities of the drivers are bypassed for NIC alias).

However NIC ALIAS is not attaching a new interface so it will not add a new MAC filter rule. In your case you need to disable MAC spoofing either way.

rsmontero:
Sorry, on a second thought we can add the IP/MAC pair in the MAC spoofing rules. So a NIC ALIAS includes IP/MAC pair and we can add that to the rules.

So for example, a VM template with:

```
NIC_ALIAS = [ NETWORK = "Public", PARENT = "private_net" ]
```

will result in:

```
NIC_ALIAS =   [ ALIAS_ID = 0, AR_ID=1, BRIDGE=virbr0, IP=1.1.1.1, MAC= 02:00:01:01:01:01...
```

So it may have sense to add that MAC to the rules. Is that your proposal?

TinoatTSI:
Yes I think that would do it. This also needs to be performed when a NIC is hot-attached.
So in both cases you would need to add the alias IP/MAC pair to the spoofing protection "whitelist".

[...]

PR #4764, opened 2020-05-19
M #-: Add IP Spoofing support for NIC_ALIAS

rdiaz-on:
Add IP addresses of a NIC_ALIAS to the corresponding ipset if the parent
NIC belongs to the "Bridged & Security Groups" network mode (fw) when a
VM is instantiated or a NIC_ALIAS attached.

Signed-off-by: Ricardo Diaz <rdiaz@opennebula.io>

### DIFF
--- a/src/vmm_mad/exec/one_vmm_exec.rb
+++ b/src/vmm_mad/exec/one_vmm_exec.rb
@@ -970,7 +970,7 @@ class ExecDriver < VirtualMachineDriver
                     ]
                 }
             ]
-        elsif nic_alias && external
+        elsif nic_alias
             steps = [
                 # Execute pre-attach networking setup
                 {
@@ -1070,6 +1070,15 @@ class ExecDriver < VirtualMachineDriver
                 {
                     :driver       => :vnm,
                     :action       => :clean
+                }
+            ]
+        elsif nic_alias
+            steps = [
+                # Execute post-boot networking setup
+                {
+                    :driver       => :vnm,
+                    :action       => :post,
+                    :parameters   => [:deploy_id]
                 }
             ]
         else
--- a/src/vnm_mad/remotes/lib/security_groups_iptables.rb
+++ b/src/vnm_mad/remotes/lib/security_groups_iptables.rb
@@ -334,6 +334,14 @@ module SGIPTables
             vars[:set_sg_out] = "#{vars[:chain]}-#{sg_id}-o"
         end
 
+        vars[:nic_aliases] = []
+
+        unless nic[:alias_ids].nil?
+            nic[:alias_ids].split(',').each do |id|
+                vars[:nic_aliases] << vm.nic_alias(id)
+            end
+        end
+
         vars
     end
 
@@ -426,6 +434,10 @@ module SGIPTables
 
             [:ip, :vrouter_ip].each do |key|
                 ipv4s << nic[key] if !nic[key].nil? && !nic[key].empty?
+                vars[:nic_aliases].each do |nic_alias|
+                    ipv4s << nic_alias[key] \
+                        if !nic_alias[key].nil? && !nic_alias[key].empty?
+                end
             end
 
             if !ipv4s.empty?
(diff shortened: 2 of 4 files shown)


src/vmm_mad/exec/one_vmm_exec.rb
```diff
@@ -1072,6 +1072,15 @@ def detach_nic(id, drv_message)
                     :action       => :clean
                 }
             ]
+        elsif nic_alias
+            steps = [
+                # Execute post-boot networking setup
+                {
+                    :driver       => :vnm,
+                    :action       => :post,
```
rsmontero, line 1080: Why not clean action?

src/vnm_mad/remotes/lib/vm.rb
```diff
@@ -121,6 +122,27 @@ def nic_build_hash(nic_element, nic)
...
+            def nics_build(xpath_filter)
+                nics = VNMNetwork::Nics.new(hypervisor)
+
+                @vm_root.elements.each(xpath_filter) do |nic_element|
+                    nic = nics.new_nic
+
+                    nic_build_hash(nic_element, nic)
+
+                    if !VNMMAD.pre_action?
+                        nic.get_info(self)
```
rsmontero, line 136: Do these functions work/make sense for NIC_ALIAS? (when calling it as nics_build('TEMPLATE/NIC_ALIAS')

src/vnm_mad/remotes/lib/vm.rb
```diff
@@ -65,6 +53,19 @@ def each_nic(block)
                 end
             end
 
+            # Get NIC_ALIAS by NIC_ID
+            #   @param element [String] the NIC_ID
+            #   @return [Hash] the NIC_ALIAS
+            def nic_alias(id)
+                if @nic_aliases
```
rsmontero, line 60: better use exit condition.

return if @nic_aliases.nil?

src/vnm_mad/remotes/lib/security_groups_iptables.rb
```diff
@@ -334,6 +334,14 @@ def self.vars(vm, nic, sg_id = nil)
             vars[:set_sg_out] = "#{vars[:chain]}-#{sg_id}-o"
         end
 
+        vars[:nic_aliases] = []
+
+        unless nic[:alias_ids].nil?
```
rsmontero, line 339: prefer if !nic[:alias_ids].nil? for single statements

src/vmm_mad/exec/one_vmm_exec.rb
rdiaz-on: Clean action delete the whole iptables chain included the rules for the parent NIC

src/vnm_mad/remotes/lib/security_groups_iptables.rb
rdiaz-on: I will change it

src/vnm_mad/remotes/lib/vm.rb
rdiaz-on: I will change it

atodorov-storpool:
Hello. 
I am sorry to step in but I am a bit disappointed with this PR. IMHO this is not `M #-` as there are two bugs related to this: #3079 and #3354 and there is already developed pull request for #3354 .
Looked at another side, besides implementing a solution that does not follow own guidelines (I mean  https://github.com/OpenNebula/one/issues/3354#issuecomment-508806696) sound like downplaying contributor's work and is not nice. At least from my point of view.

Such behaviour and delaying PRs with years is highly discouraging to do project contributions.

Regards,
Anton

atodorov-storpool:
^ @rsmontero 
https://github.com/OpenNebula/one/pull/4764#issuecomment-645843393

rsmontero:
Thanks @atodorov-storpool for this an other contributions to the project. As you see this PR is still in review and will include reference to the affected bugs. 

As per the implementation this is indeed inspired by your PR, but re-written to improve its design and better fit in the driver. Also note that design decisions evolve with the code and need to consider its relation with other new components. Finally, we need to prioritize our available development resources, that unfortunately force us to delay some contributions. Please note that these decisions are not easy for us, and by not means imply that we underestimate your contributions.

Thanks again for your valuable contributions and hope you'll keep making them in the future.

mzealey:
What is the status of this? It would be nice to have this included in opennebula...

mzealey:
I have tested that this commit applies fine onto 5.12.0.3 and fixes the issue by adding/removing alias ips to the ipset.

Issue #4734, opened 2020-05-14
Docker Machine driver breaking change


feldsam:
**Description**
PR #4385 introduced a breaking change in the docker-machine driver. After an update, all existing docker-machines are unusable.

**To Reproduce**
Update driver on a system with existing machines.

**Expected behavior**
After an update, existing machines have to work. We need to implement a fallback for old VM identification.

**Details**
 - Affected Component: docker_machine
 - Hypervisor: any
 - Version: 5.10.4+

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [x] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

feldsam:
I am proposing to add temporary falback using VM name until next major release, 5.12 or 5.14, and then remove it. So users will have time to recreate docker machines.

carloshm91:
Hello @feldsam thanks for you interest in the docker-machine driver for OpenNebula.

We have the fix for this https://github.com/OpenNebula/one/pull/4789.

Test it if you want inside your environment and let us know how was.

feldsam:
Hello @carloshm91 , thanks! I have just one comment about your code, check it. I will compile myself this updated version and test it

carloshm91:
Hello @feldsam where do you put the comment? Yes, please test it and let us know.

feldsam:
Hello @carloshm91, I wrote it on your PR

feldsam:
Just tested, looks like working well. Tested with my changes https://github.com/OpenNebula/one/pull/4797

tinova:
Thanks for this quality feedback

PR #4789, opened 2020-05-21
B #4734: fix driver breaking change

carloshm91:
-> master

### DIFF
--- a/src/docker_machine/src/docker_machine/opennebula.go
+++ b/src/docker_machine/src/docker_machine/opennebula.go
@@ -525,7 +525,7 @@ func (d *Driver) GetIP() (string, error) {
 	if d.MachineId == 0 {
 		vm_id, err2 = controller.VMs().ByName(d.MachineName)
 		if err2 != nil {
-			return err2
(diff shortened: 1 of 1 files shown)


src/docker_machine/src/docker_machine/opennebula.go
```diff
@@ -519,7 +519,20 @@ func (d *Driver) GetURL() (string, error) {
...
+	var err2 error
+	var vm_id int
+
+	if d.MachineId == 0 {
+		vm_id, err2 = controller.VMs().ByName(d.MachineName)
+		if err2 != nil {
+			return "", err2
+		}
+		d.MachineId = vm_id
+	} else {
```
feldsam, line 531: why such complicated construct? what about this:

```
        if d.MachineId == 0 {
 		vm_id, err2 = controller.VMs().ByName(d.MachineName)
 		if err2 != nil {
 			return "", err2
 		}
 		d.MachineId = vm_id
 	}

        vm, err := controller.VM(d.MachineId).Info(false)
```

Issue #4793, opened 2020-05-22
VM disk monitoring not working as expected


atodorov-storpool:
**Description**
After enabling the workaround of #4791 something is working but it looks like the sent monitoring data from the remote hosts is not processed, despite the `Successfully monitored VM: $VM_ID` log message

**To Reproduce**
install opennebula and a remote KVM host
instantiate a VM in the default ssh SYSTEM datatore
check the VM disk size

**Expected behavior**
The VM disks size should be updated

**Details**
 - Affected Component: [Monitring]
 - Hypervisor: [KVM]
 - Version: [development]

**Additional context**
As far as I understand there are two ways to feed VM disk stats to OpenNebula:
1. on the frontend node - TM_MAD/monitor when called by for SYSTEM datastore which is not tagged for "at the host" monitoring (Cant twist my mind to name it "tagged for local monitoring"...)
2. On the hosts - TM_MAD/monitor_ds called by the remote monitoring probe

the monitoring message should be 
```
VM = [ ID = $VM_ID, MONITOR = "<base64>
DISK_SIZE=[ID=$DISK_ID,SIZE=$DISK_SIZE]
SNAPSHOT_SIZE = [ID=$SNAP_ID, DISK_ID=$DISK_ID, SIZE=$SNAP_SIZE]
</base64>" ]
```
The message is logged as successfully monitored by monitors:
```
Fri May 22 10:41:02 2020 [Z0][HMM][I]: Successfully monitored VM: 3
```
But in the VMs the disks sizes are 0:

```
[root@one512 ssh]# onevm show 3
VIRTUAL MACHINE 3 INFORMATION                                                   
ID                  : 3                   
NAME                : Ttylinux - KVM-3    
USER                : oneadmin            
GROUP               : oneadmin            
STATE               : ACTIVE              
LCM_STATE           : RUNNING             
LOCK                : None                
RESCHED             : No                  
HOST                : vs04                
CLUSTER ID          : 0                   
CLUSTER             : default             
START TIME          : 05/20 16:15:55      
END TIME            : -                   
DEPLOY ID           : 6d8f347e-fbce-4d26-9df5-d455d6a2c11b

VIRTUAL MACHINE MONITORING                                                      
CPU                 : 1.01                
MEMORY              : 106.1M              
NETTX               : 0K                  
NETRX               : 0K                  
DISKRDBYTES         : 13855320            
DISKRDIOPS          : 974                 
DISKWRBYTES         : 596992              
[...]

PR #4820, opened 2020-05-27
B #4793: Update tm_mad/shared/monitor

atodorov-storpool:
MONITOR_ADDRESS could differ from _127.0.0.1_ in _/etc/monitord.conf_
The default netcat (NC) on centos7.8 is _nmap-ncat_ which do not accept '0' for _--wait_
(_Ncat: Invalid -w timeout (must be greater than 0). QUITTING._)

The proposed change defines them as shell variables, not sure where to configure them to match the configuration in /etc/monitord.conf, though.

Also, needs a dependency on 'nc' in the packaging...


atodorov-storpool:
fixed better in upstream

PR #4824, opened 2020-05-27
B #4793: Read monitor address from monitord.conf

paczerny:


paczerny:
Do not merge, although it covers some cases, it's still not complete

PR #4847, opened 2020-06-01
B #4793: Fix VM disks monitoring for LVM

paczerny:


xorel:
Can you change all variables to the upper-case?
Also `LVM_SIZE_CMD` seems unused (although it was there before)

paczerny:
So the preferred style in OpenNebula is to use upper-case?
I saw recommendations to use lower case https://stackoverflow.com/questions/32144204/what-is-a-convention-for-naming-a-constant-in-bash
In other OpenNebula scripts the style is mixed https://github.com/OpenNebula/one/blob/master/src/mad/sh/scripts_common.sh#L330
In this script I used the same convention as nearby code

xorel:
I think upper case is mainly used, mixing the case within the short file seems ugly to me.

PR #4870, opened 2020-06-02
M #-: add default sqlite3 database path

al3xhh:


src/onedb/onedb.rb
```diff
@@ -33,6 +33,8 @@ def initialize(ops)
         end
 
         if ops[:backend] == :sqlite
+            ops[:sqlite] = '/var/lib/one/one.db' if ops[:sqlite].nil?
```
rsmontero, line 36: Also needs to be set for self-container installations

PR #4884, opened 2020-06-04
M #-: OneProvision add firecracker support

al3xhh:


share/oneprovision/ansible/roles/opennebula-node-firecracker/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+- name: Install OpenNebula node Firecracker package
+  package:
+    name: opennebula-node-firecracker
+    state: latest
+  register: result
+  until: result is success
+  retries: 3
+  delay: 10
+
+- name: Install qemu-img
```
vholer, line 12: Why explicitly install qemu-img / qemu-utils?

@christian7007:
- what parts require them?
- how are these packages installed on testing then?
- if needed, they must be set as `opennebula-node-firecracker` package dependency
(otherwise having them in Ansible role here is only a workaround)

Issue #5000, opened 2020-07-27
The VM RAW data can't have several <devices /> entries


baby-gnu:
**Description**

We upgraded our OpenNebula to 5.12.1 successfully (with minor documentation issues [#4998](https://github.com/OpenNebula/one/issues/4998) and [#4999](https://github.com/OpenNebula/one/issues/4999)).

There is a regression in the hanling of `RAW` data.

We build some template with several devices `RAW` data like:

```
RAW=[
  DATA="	<devices>
		<controller type='scsi' index='0' model='virtio-scsi'>
			<driver queues='2'/>
		</controller>
	</devices>
	<devices>
		<video>
			<model type='vga' vram='16384' heads='1'/>
		</video>
	</devices>",
  TYPE="kvm" ]
```

it was working fine with `5.10.1` but fails with `5.12`:

```
Mon Jul 27 12:14:24 2020 [Z0][ONE][E]: Invalid RAW section: cannot validate DATA with domain.rng schema
Mon Jul 27 12:14:24 2020 [Z0][ReM][E]: Req:2304 UID:4 one.vm.allocate result FAILURE [one.vm.allocate] Invalid RAW section: cannot validate DATA with domain.rng schema
```

When I remove one `<devices />` element, it's working.

Having several `<devices />` entries is not a problem for `libvirt` since OpenNebula generate itself `deployment.0` files containing several ones:

``` xml
<domain type='kvm' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
	<name>one-388035</name>
	<title>etb1.horus</title>
	<vcpu><![CDATA[4]]></vcpu>
	<cputune>
		<shares>205</shares>
	</cputune>
	<memory>3145728</memory>
	<os>
		<type arch='x86_64'>hvm</type>
	</os>
	<devices>
		<emulator><![CDATA[/usr/bin/qemu-system-x86_64]]></emulator>
		<disk type='file' device='disk'>
			<source file='/var/lib/one//datastores/0/388035/disk.0'/>
			<target dev='sda'/>
			<boot order='1'/>
			<driver name='qemu' type='qcow2' cache='none' discard='unmap'/>
			<address type='drive' controller='0' bus='0' target='0' unit='0'/>
		</disk>
		<disk type='file' device='disk'>
			<source file='/var/lib/one//datastores/0/388035/disk.1'/>
			<target dev='sdb'/>
			<driver name='qemu' type='qcow2' cache='none' discard='unmap'/>
[...]

PR #5001, opened 2020-07-27
B #5000: RAW data should accept several <devices /> entries

baby-gnu:
* share/schemas/libvirt/domaincommon.rng: make the element `devices` a `zeroOrMore` element.


baby-gnu:
This should be fixed directly in [libvirt](https://gitlab.com/libvirt/libvirt/-/tree/master/docs/schemas).

PR #5067, opened 2020-09-08
Allow disk resize with snapshots

chiraganand:


chiraganand:
Closing this PR because of failed checks, will open another.

Issue #4047, opened 2019-12-12
FAILURE [one.vm.diskresize] Cannot resize a disk with snapshots


mzealey:
**Description**

https://github.com/OpenNebula/one/blob/8d12e78fbe15ba43aa53f098da294a32e6e2cc85/src/rm/RequestManagerVirtualMachine.cc#L3253 stops you from being able to resize a vm disk when it has any snapshots. This is not a limitation on all storage types. Please change this so that a driver can specify if it will allow resizing when there are snapshots.

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

cbr46:
Hello,
do you have a status update on this problem?

Issue #1260, opened 2017-11-20
Improve snapshots management


OpenNebulaProject:
---

Author Name: **Vlastimil Holer** (Vlastimil Holer)
Original Redmine Issue: 5223, https://dev.opennebula.org/issues/5223
Original Date: 2017-07-11

---

https://dev.opennebula.org/issues/4393#note-17

1. even with ALLOW_ORPHANS="yes" in TM_MAD conf it is not possible to delete any disk snapshot
1.1 Sunstone don't allow deleting "current" snapshot with reason "[one.vm.disksnapshotdelete] Cannot delete the active snapshot" 
1.2 Sunstone don't allow deleting "parent" snapshot with reason "[one.vm.disksnapshotdelete] Cannot delete snapshot with children" 
1.3 it is not allowed to resize a disk if it has snapshots with reason "[one.vm.diskresize] Cannot resize a disk with snapshots"

(1.4 [new feture?] there is no option to do (atomic)snapshots on all VM disks at once in a single transaction)

All of the above constraints are not issue in StorPool. The snapshots in StorPool are exactly how the snapshot tree is drawn. And the active disk volume is the final leaf to a snapshot then. Each snapshot keeps only the changes against its parent. When a snapshot is deleted its changes are merged to its leaf(s) - either snapshot or the active volume. I'll try to explain with a simple drawing

a VM has one disk named DISK_0, when you do a snapshot it will have

```
SNAPSHOT_A
     \
      DISK_0

```

When you take a second snapshot you'll have

```
SNAPSHOT_A
     \
      SNAPSHOT_B
           \
            DISK_0

```

When you delete SNAPSHOT_A the shared data is copied to ites leaf.. in this case SNAPSHOT_B
When SNAPSHOT_B is deleted the shared data is copied back to its leaf - DISK_0
The only thing I can think of is that when a disk snapshot is created the actual disk size should be saved as an attribute to the snapshot and when a snapshot is restored - to alter the value in the VM tample with the value saved in the snapshot.
For example if we have a disk with size 10GB, take a snapshot and then resize the disk to 15 GB. When you revert the disk to the snapshot the resulting disk size will be reverted back to 10GB.

I am almost sure same is with CEPH too.

Regarding the VM snapshots. 
[...]

PR #5068, opened 2020-09-08
B #4047: allow resize for disks with snapshots

chiraganand:
This change has been tested on Linstor datastore using linstor_un
driver (master). It makes the datastore driver decide whether it
allows resizing of disks which have snapshots. Also, refs #1260.


chiraganand:
Closing this PR because of failed checks, will open another.

PR #5070, opened 2020-09-09
B #4047: allow resize for disks with snapshots

chiraganand:
This change has been tested on Linstor datastore using linstor_un
driver (master). It makes the datastore driver decide whether it
allows resizing of disks which have snapshots. Also, refs #1260.

### DIFF
--- a/src/rm/RequestManagerVirtualMachine.cc
+++ b/src/rm/RequestManagerVirtualMachine.cc
@@ -3248,15 +3248,6 @@ void VirtualMachineDiskResize::request_execute(
         return;
     }
 
-    if ( disk->has_snapshots() )
-    {
-        att.resp_msg = "Cannot resize a disk with snapshots";
-        failure_response(ACTION, att);
-
-        vm->unlock();
-
-        return;
-    }
 
     /* ------------- Get information about the disk and image --------------- */
     bool img_ds_quota, vm_ds_quota;


src/rm/RequestManagerVirtualMachine.cc
```diff
@@ -3248,15 +3248,6 @@ void VirtualMachineDiskResize::request_execute(
         return;
     }
 
-    if ( disk->has_snapshots() )
```
rsmontero, line 3251: Probably it'd be better to let the driver choose at configuration level. Similar to the migration capabilities https://github.com/OpenNebula/one/blob/26ce7484dec1b305ead4a188056b8b145c84e891/src/rm/RequestManagerVirtualMachine.cc#L308

We'll need to:
* Add an option to TM_MAD_CONF attribute in oned.conf ALLOW_SNAP_RESIZE , for example. 
* Let drivers like linstor that supports the feature to set this to true
* The removed code should check for the ds attribute to grant or deny the resize operation.

chiraganand: > We'll need to:
> 
> * Add an option to TM_MAD_CONF attribute in oned.conf ALLOW_SNAP_RESIZE , for example.
> * Let drivers like linstor that supports the feature to set this to true
> * The removed code should check for the ds attribute to grant or deny the resize operation.

Sure, I will work on these and send another review.

rsmontero: Awesome!

A couple of pointers. 

1. We need to add an attribute in the Datastore (system_ds) that states if you can resize snapshots of a VM disk. For example DS_MIGRATE is handled here: https://github.com/OpenNebula/one/blob/master/src/datastore/Datastore.cc#L348. This will add the ALLOW_SNAP_RESIZE to the Datastore template.

2. Use the new attribute to check if resize is allowed for the VM. Probably as the VM is already running in a SYSTEM_DS, you can use https://github.com/OpenNebula/one/blob/master/include/VirtualMachine.h#L472 to get the Datastore ID, and with that ID get the Datastore object and the configured ALLOW_SNAP_RESIZE value.

3. Fallback to not allow for compatibility

PR #5099, opened 2020-09-21
F #5102: Show stats of IP

bashsiz:
with this script you can check the state of an IP and also you can get a free IP from your networks

### DIFF
--- a/src/cli/oneip
+++ b/src/cli/oneip
@@ -1,0 +1,163 @@
+#!/bin/bash
+
+# -------------------------------------------------------------------------- #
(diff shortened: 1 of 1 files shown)


tinova:
thanks, this is a wonderful contribution :)

would you mind adding it to https://github.com/tinova/one-tools instead? I think these kind of tools makes more sense in that complementary repo.

bashsiz:
> thanks, this is a wonderful contribution :)
> 
> would you mind adding it to https://github.com/tinova/one-tools instead? I think these kind of tools makes more sense in that complementary repo.

Thank you
OK

Issue #5196, opened 2020-12-04
monitord can't insert monitoring data 


atodorov-storpool:
**Description**

Having the following in monitord.log between the successful :

```
...
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 7
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 8
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 5
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 3
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 9
Fri Dec  4 18:17:22 2020 [Z0][HMM][I]: Successfully monitored VM: 12
Fri Dec  4 18:17:26 2020 [Z0][ONE][E]: SQL command was: REPLACE INTO vm_monitoring (vmid, last_poll, body) VALUES (4,1607098646,'<MONITORING><CPU><![CDATA[30.78]]></CPU><DISKRDBYTES><![CDATA[796155776]]></DISKRDBYTES><DISKRDIOPS><![CDATA[27270]]></DISKRDIOPS><DISKWRBYTES><![CDATA[5414258176]]></DISKWRBYTES><DISKWRIOPS><![CDATA[883527]]></DISKWRIOPS><ID><![CDATA[4]]></ID><MEMORY><![CDATA[16849780]]></MEMORY><NETRX><![CDATA[15289195964]]></NETRX><NETTX><![CDATA[31285031216]]></NETTX><TIMESTAMP><![CDATA[1607098646]]></TIMESTAMP></MONITORING>'), error 1046 : No database selected
Fri Dec  4 18:17:26 2020 [Z0][HMM][E]: Unable to write monitoring to DB
Fri Dec  4 18:17:26 2020 [Z0][HMM][I]: Successfully monitored VM: 5
Fri Dec  4 18:17:26 2020 [Z0][HMM][I]: Successfully monitored VM: 6
Fri Dec  4 18:17:26 2020 [Z0][HMM][I]: Successfully monitored VM: 7
[...]

PR #5198, opened 2020-12-07
B #5196: fix mysql reconnect

atodorov-storpool:
set charset with mysql_option() 
add the database name to mysql_real_connect()
set flag to remember the options in mysql_real_connect()
append th the db_connect array in a single loop


paczerny (approved):
I wasn't able to reproduce the "error 1046 : No database selected".
At least this commit is more elegant way to init the DB connection.

rsmontero (approved):

Issue #5258, opened 2021-02-18
Onegate ruby client does not terminate a VM (nor other similar operations) 


rcampuz:
**Description**
Using the onegate ruby client from inside a VM, the "terminate" action does not return any error and it does not produce any effect.

**To Reproduce**
Create a service with 2 VMs (using Ubuntu 20.04 image) --> VM ids = 52, 53
ssh in one of the VMs (the one with id=52)
Validate that onegate works fine executing any random command: onegate vm show 53 -j
Execute the action to terminate the VM with id=53: onegate vm terminate 53

**Expected behavior**
The VM with id=53 gets terminated

**Details**
 - Affected Component: Onegate
 - Hypervisor: KVM
 - Version: 5.4.12.03

**Additional context**

## Progress Status
- [x] Branch created 
- [x] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rcampuz:
I found where the problem is, I will try to commit the code myself.

It is here: one/share/onegate/onegate line 676
`    when "resume",
         "stop",
         "suspend",
         "terminate",
         "reboot",
         "poweroff",
         "resched",
         "unresched",
         "hold",
         "release",
         # Compatibility with 4.x
         "delete",
         "shutdown", # <-- line 676, the comma has to be removed
        if ARGV[2]
            action_hash = {
                "action" => {
                    "perform" => ARGV[1]
                }
            }`

I tested it removing that "comma" and now it works.

rcampuz:
The comma was removed from the code.
Pull request: [https://github.com/OpenNebula/one/pull/5259]

Tested in our OpenNebula instance:
onegate vm terminate 53 (executed from VM with id=52 now works)

PR #5264, opened 2021-02-23
B #5258: Onegate ruby client do not terminate a VM (nor other similar…

rcampuz:
… operations)


al3xhh:
Already merged, thanks!

Issue #5301, opened 2021-03-16
Docker-machine driver try to resume VM when it is in Pending state


feldsam:
Hello, I see error logs like this

```
Req:2720 UID:3 one.vm.action result FAILURE [one.vm.action] Error performing action "resume": This action is not available for state PENDING
```

I think it is related to following line:
https://github.com/OpenNebula/one/blob/051b36e0d91746d8e7ffc77cfb365d36e337ec9e/src/docker_machine/src/docker_machine/opennebula.go#L693

What is purpose of resume action?

feldsam:
PR #5303

al3xhh:
Changes are already merged, closing this, reopen it if needed.

PR #5303, opened 2021-03-16
B #5301: docker-machine enhancements

feldsam:
- fixed issue with resume vm in pending state
- enhanced GetIP method to minimize API calls
- GetURL  return address only if machine is running
- code cleanup

Signed-off-by: Kristian Feldsam <feldsam@gmail.com>

### DIFF
--- a/src/docker_machine/src/docker_machine/opennebula.go
+++ b/src/docker_machine/src/docker_machine/opennebula.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"strconv"
 	"time"
 
@@ -46,7 +47,6 @@ type Driver struct {
 }
 
 const (
-	defaultTimeout      = 1 * time.Second
 	defaultSSHUser      = "docker"
 	defaultCPU          = "1"
 	defaultVCPU         = "1"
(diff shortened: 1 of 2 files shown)


src/docker_machine/src/docker_machine/opennebula.go
```diff
@@ -676,22 +678,25 @@ func (d *Driver) GetState() (state.State, error) {
...
-	var vm_id int
 
-	if d.MachineId == 0 {
-		vm_id, err = controller.VMs().ByName(d.MachineName)
-		if err != nil {
-			return err
-		}
-		d.MachineId = vm_id
-	}
+	if drivers.MachineInState(d, state.Stopped)() {
```
christian7007, line 683: Probably this should take care of resuming also VMs in `Saved` state (i.e `STOPPED` and `SUSPENDED`). What do you think?

christian7007 (commented):
Hello @feldsam, sorry for the late review, we've been a bit busy. I've just added a comment to one of the changes.

Thanks for your contribution!

src/docker_machine/src/docker_machine/opennebula.go
feldsam: Hi, I didn't implement other states, because the docker-machine has only methods start, stop, and restart. So VM should not be in the Saved state.

christian7007: Ok, it makes sense, thanks!

PR #5306, opened 2021-03-22
Unable to set :memory_dumps to false at vCenter Driver

slntopp:
Hi, recently we were trying to set `memory_dumps` to `false` for snapshots creation using `/var/lib/one/remotes/etc/vmm/vcenter/vcenterrc`. But found out it doesn't make any effect. And more than that method definition has following lines at the beginning:
```ruby
memory_dumps = true
memory_dumps = CONFIG[:memory_dumps] if CONFIG[:memory_dumps]
```
Which makes no sense, because config value is going to be used only if its `true`.
If that's was `.nil?` check, then maybe my edit would fit.


tinova:
thanks!! this is fixed in 6.0 a37ab8604766b46f05fce1dc417b7c5e99a9d05c

Issue #5382, opened 2021-05-05
Firecracker ignores INITRD


sw37th:
**Description**

On Firecracker API, `BootSource` descripter has `initrd_path` property that specifies the path to the initrd image file. It's used when the driver module required for booting, such as virtio_blk, is not built into the kernel.

* https://github.com/firecracker-microvm/firecracker/blob/main/src/api_server/swagger/firecracker.yaml

OpenNebula can deproy the initrd file from Datastore to the host using the `INITRD_DS` attribute, but Firecracker VMM driver ignores it and creates the deployment.file that lacks the `initrd_path` property.

**To Reproduce**

Create the initrd Image on files datastore:

```
oneadmin@opennebula:~$ oneimage list --no-expand
  ID USER     GROUP    NAME            DATASTORE     SIZE TYPE PER STAT RVMS
   2 oneadmin oneadmin focal-initrd    files          79M RD    No rdy     0
   1 oneadmin oneadmin focal-kernel    files          43M KL    No rdy     0
   0 oneadmin oneadmin focal-rootfs    default      1024M OS    No rdy     0
```

Create a VM template with the following OS data:

```
OS=[
  INITRD_DS="$FILE[IMAGE_Id=2]",
  KERNEL_CMD="console=ttyS0 reboot=k panic=1 pci=off i8042.noaux i8042.nomux i8042.nopnp i8042.dumbkbd",
  KERNEL_DS="$FILE[IMAGE_ID=1]" ]
```

Instantiate a MicroVM. The initrd file is deployed to the host:

```
oneadmin@host003:~$ ls -l ~/datastores/0/13/initrd
-rw-r--r-- 1 oneadmin oneadmin 82813842 May  4 19:50 /var/lib/one/datastores/0/13/initrd

oneadmin@host003:~$ xmllint --xpath //TEMPLATE/OS/INITRD ~/datastores/0/13/vm.xml
<INITRD><![CDATA[/var/lib/one//datastores/0/13/initrd]]></INITRD>
```

But the deployment.file does not contain the `initrd_path` property.

```
oneadmin@host003:~$ jq '."boot-source"' ~/datastores/0/13/deployment.file
{
  "kernel_image_path": "kernel",
  "boot_args": "console=ttyS0 reboot=k panic=1 pci=off i8042.noaux i8042.nomux i8042.nopnp i8042.dumbkbd"
}
```

**Expected behavior**

The deployment.file contains `initrd_path` property that specifies the path to the initrd file.

[...]

PR #5383, opened 2021-05-05
B #5382: Fix Firecracker ignores INITRD

sw37th:
Check the INITRD attribute in the XML and add initrd_path property to boot-source if it exists.


rsmontero:
merge as per 8e9c2f4d1fae64fee4a0b3d2899c46d70a74eb51

rsmontero:
THANKS!!!!!!

Issue #5442, opened 2021-06-30
VM networks not working with CentOS 8.4+Advanced Virtualization


atodorov-storpool:
**Description**
The Advanced Virtualization SIG for CentOS 8.4 provides libvirt 7.0.0 that have a different output when creating a domain from the cmd line with virsh.

**To Reproduce**
Setup OpenNebula KVM with CentOS8.4 + Advanced Virtualization
Instantiate a VM (with spoofing filters, think not important)
The VM and its networking is not working - no iptables rules are created

**Expected behavior**
The VM and its networking should work

**Details**
 - Affected Component: [VMM_MAD]
 - Hypervisor: [KVM]
 - Version: [5.12.10-EE]

**Additional context**
The root cause is that OpenNebula relies on a fuzzy/buggy parsing the output from an external program (libvirt/virsh) for data that is available in the deployment XML.

Patch pending to get the ID of the bug...

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

atodorov-storpool:
Looks like the issue is fixed in 6.0.x but not in the current 5.12 LTS?

rsmontero:
Thanks for the heads up, merged into one-5.12 ee

PR #5443, opened 2021-06-30
B #5442: get DOMAIN_ID from the deployment file

atodorov-storpool:
and optimize domain UUID extraction with awk

### DIFF
--- a/src/vmm_mad/remotes/kvm/deploy
+++ b/src/vmm_mad/remotes/kvm/deploy
@@ -34,8 +34,8 @@ DATA=`virsh --connect $LIBVIRT_URI create $DEP_FILE`
 
 if [ "x$?" = "x0" ]; then
 
-    DOMAIN_ID=$(echo $DATA | sed 's/Domain //' | sed 's/ created from .*$//' | tr -d "'")
-    UUID=$(virsh --connect $LIBVIRT_URI dominfo $DOMAIN_ID | grep UUID: | awk '{print $2}')
+    DOMAIN_ID=$(xmllint --xpath '/domain/name/text()' "$DEP_FILE")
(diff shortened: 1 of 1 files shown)


src/vmm_mad/remotes/kvm/deploy
```diff
@@ -34,8 +34,8 @@ DATA=`virsh --connect $LIBVIRT_URI create $DEP_FILE`
 
 if [ "x$?" = "x0" ]; then
 
-    DOMAIN_ID=$(echo $DATA | sed 's/Domain //' | sed 's/ created from .*$//' | tr -d "'")
-    UUID=$(virsh --connect $LIBVIRT_URI dominfo $DOMAIN_ID | grep UUID: | awk '{print $2}')
+    DOMAIN_ID=$(xmllint --xpath '/domain/name/text()' "$DEP_FILE")
```
xorel, line 37: This would work but strictly speaking xmllint (xmllint2/libxml2-utils) is not in the packages dependency, so far we only have libxml2-dev which itself requires xmllint. Anyway, what about using single sed command like:
```
sed "s/Domain '\{0,1\}\([^ ']*\)'\{0,1\} created from deployment.*/\1/"
```
Should be faster than parsing the deployment file or is it too cryptic?

atodorov-storpool: My point is that _anything_ parsing a human-readable text is prone to fail because the text could change over the time, for example, due to a developer decision, enabled translation, etc...
IMO the domain XML is much more _stable_.

xorel: ok, good point, I agree with the patch

xorel (approved):

rsmontero:
Merged as per 1a041abb6fb6d36e7564453fdddf561abfa6b7c8

THANKS

PR #5497, opened 2021-08-17
M #-: better English wording

baby-gnu:
I'm not an English expert, neither a native speaker but a comma `,` before `which` hurts me a lot.
In the opposite, the comma should be present to separate the condition to the action.


rsmontero:
THANKS merged as per b186d196a69802f9eca5b8fbaf070c77e07dd342

PR #5514, opened 2021-08-29
Stop at the first match for LDAP match_user_regex option

### LINKED ISSUE #5513: Checking of LDAP match_user_regex is done with modified value
**Description**
The `match_user_regex` option is able to modify the username, but after one of them matches it then uses the modified username to check the others.

**To Reproduce**
If I have the following ldap auth configuration:
```
:match_user_regex:
  "^(.+)!$": admin
  "^([^!]+)$": user
```

Then an admin user `example!` is matched against the first regex, and then the username to lookup is changed to `example`. It then matches the next regex too. This should not happen.

**Expected behavior**
Only one regex can match and the modified username is not used for any other regex.

**Details**
 - Affected Component: 
 - Version: 6.0.0.2

## Progress Status
- [ ] Branch created 
- [ ] Testing - QA
- [ ] Documentation 

nomis:
When a match is made, the username is changed. It's then inappropriate to attempt further matches because we're not using the original username.

If multiple matches were made with the original username, there could be no single modified username to return. An alternative would be to return pairs of server and modified username.

Fixes #5513.

### DIFF
--- a/src/authm_mad/remotes/ldap/ldap_auth.rb
+++ b/src/authm_mad/remotes/ldap/ldap_auth.rb
@@ -274,6 +274,7 @@ def get_server_order(opts, user)
                 user = m[1] if m[1]
 
                 order << to_array(server)
+                break
             end
         end
 


xorel (approved):

xorel:
Thanks for the bugfix!

rsmontero (approved):

Issue #5530, opened 2021-09-02
VM snapshot lost after migration


rene-bayer:
**Description**
a vm snapshot (not disk snapshot) is broken/lost after live migration or power-off

**To Reproduce**

- create  a new virtual machine named test1 and start it
- open the virtual machine in sunstone, navigate to "Snapshots" and create one
- wait for the snapshot to finish
- live migrate vm "test1" to an other hypervisor
- the snapshot is still listed in sunstone, but with `virsh snapshot-list <domain>` we see that the snapshot is gone
- there's no way to remove that broken snapshot in sunstone anymore

- create a new virtual machine named test2 and start it
- open the virtual machine in sunstone, navigate to "Snapshots" and create one
- wait for the snapshot to finish
- power off vm "test2"
- snapshot in sunstone is gone
- power on vm "test2"
- snapshot in sunstone is still gone, but with `virsh snapshot-list <domain>` you can still see the snapshot

**Expected behavior**
In both cases I would expect the snapshot to be still there and correctly shown in Sunstone

**Details**
 - Affected Component: Sunstone, Storage, libvirtd
 - Hypervisor: KVM
 - Storage: shared fs - TM_MAD="qcow2"
 - Version: 6.0.2

**Additional context**
NaN

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

atodorov-storpool:
duplicate of https://github.com/OpenNebula/one/issues/5444

atodorov-storpool:
The snapshots created by opennebula are of type internal and they are discarded on domain poweroff and migrate

DarkSunOne:
Okay, but even then, the current behavior is wrong...
So the expected behavior should be that the snapshot is cleared everywhere in every situation, on opennebula side and libviert side.

christian7007:
Hi @DarkSunOne,

Regarding the problem when the VM is powered off. Can you confirm that you have the `KEEP_SNAPSHOTS = "yes"` in the KVM `VM_MAD` section of your oned.conf. If not, could you set it and try that case again?

DarkSunOne:
Hi @christian7007 
sry took a little ...

[...]

PR #5583, opened 2021-10-19
F #5530: Redefine snapshots after live migration

rene-bayer:
According to issue  #5530 I implemented the recommend workaround for live migrations from [here](https://listman.redhat.com/archives/libvirt-users/2013-March/msg00117.html)

I already tested this patch internally.
VM Snapshots survived a live migration (tested revert and delete afterwards)
I also added the redefinition of the current snapshot.


rsmontero:
Merged as per 5dfe830feed89debd0e3cfd0ca9e01af26252e68

Issue #5483, opened 2021-08-01
Install One CLI Tools on Mac OS


Th0masL:
**Description**
It would be great to be able to install the CLI Tools on Mac OS.

**Use case**
I want to use some of the CLI binaries to execute API calls against the OpenNebula API.
I'm using a Mac laptop and it's annoying to have to connect to a Linux machine in order to be able to use the CLI Tools.

**Changes**
Is there a way to build the CLI Tools only, from source, on Mac OS ?
I saw that the CLI binaries were ruby code/gems, so I think it should be doable.
If so, could someone update the documentation to explain how to install it ?
If it's not possible yet to install those on Mac OS, how complicated would that be to implement such functionality ?

## Progress Status
- [ ] Branch created
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
The whole CLI is written in Ruby, you just need to do the following:

- Install Ruby.
- Install all the dependencies gems.
- Install the CLI code and the configuration files.
- Run the commands against the OpenNebula instance running in a different server using `--endpoint` parameter.

Closing this.

danhigham:
@al3xhh The install script (install.sh), in it's current state, does not work on Mac OS due to a subtle difference in the way "cp" works on linux and Mac OS, see my PR for a fix (https://github.com/OpenNebula/one/pull/5136)

There are also three dependencies that have to be installed to satisfy some gems with native extensions (augeas, postgresql and mysql), these should also be installed using homebrew first before installing the gems. It also might be worth pointing out that the version of Ruby that ships with Mac OS will not suffice due to dependencies on openssl, a better path would be to install stable Ruby 2.7 with rbenv first. I would like to contribute to the documentation for installing the CLI on Mac, where would be the best place to do this?

PR #5602, opened 2021-11-05
B #5483: Add 'flow' folder to LIB_OCA_CLIENT_DIRS

danhigham:
There is a an issue in install.sh when run on Mac OS, instead of creating the folder 'src/oca/ruby/opennebula/flow', the files that follow in the RUBY_OPENNEBULA_LIB_FLOW_FILES list get written in turn to a FILE with the path 'src/oca/ruby/opennebula/flow'. This is due to a subtle difference in the way the 'cp' command works in linux and Mac OS.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -540,7 +540,8 @@ ONEFLOW_DIRS="$ONEFLOW_LOCATION/lib \
               $ONEFLOW_LOCATION/lib/models"
 
 LIB_OCA_CLIENT_DIRS="$LIB_LOCATION/ruby \
-                 $LIB_LOCATION/ruby/opennebula"
+                 $LIB_LOCATION/ruby/opennebula \
+                 $LIB_LOCATION/ruby/opennebula/flow"
 
 LIB_CLI_CLIENT_DIRS="$LIB_LOCATION/ruby/cli \
                      $LIB_LOCATION/ruby/cli/one_helper"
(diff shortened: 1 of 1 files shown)


al3xhh (changes requested):
Why do we need this change?

We have that directory added here:

```
  LIB_DIRS="$LIB_LOCATION/ruby \                                                  
            $LIB_LOCATION/ruby/opennebula \                                       
            $LIB_LOCATION/ruby/opennebula/flow \ ```

**Comment by danhigham** [comment]
> Why do we need this change?
> 
> We have that directory added here:
> 
> ```
>   LIB_DIRS="$LIB_LOCATION/ruby \                                                  
>             $LIB_LOCATION/ruby/opennebula \                                       
>             $LIB_LOCATION/ruby/opennebula/flow \ ```
> ```

The use case here is if you are only installing the CLI using the install script by using the -c flag

**Comment by al3xhh** [comment]
Thanks for the comment!

But instead of adding the directory where you added it, I prefer this:

```
diff --git a/install.sh b/install.sh
index a3d0da4411..d4b2a7f602 100755
--- a/install.sh
+++ b/install.sh
@@ -540,7 +540,8 @@ ONEFLOW_DIRS="$ONEFLOW_LOCATION/lib \
               $ONEFLOW_LOCATION/lib/models"
 
 LIB_OCA_CLIENT_DIRS="$LIB_LOCATION/ruby \
-                 $LIB_LOCATION/ruby/opennebula"
+                     $LIB_LOCATION/ruby/opennebula \
+                     $LIB_LOCATION/ruby/opennebula/flow"
 
 LIB_CLI_CLIENT_DIRS="$LIB_LOCATION/ruby/cli \
                      $LIB_LOCATION/ruby/cli/one_helper"
```

Could you please update the PR so we can merge it?

Thanks!!!!

danhigham:
> LIB_OCA_CLIENT_DIRS

Ah! That _does_ seem much cleaner! I changed the PR, as requested.

al3xhh (approved):

rsmontero:
Merge as per 07be714219af3e448e10b09d336831176da4804b

PR #5675, opened 2022-01-02
F #5688: add request context to client calls

peterwillis:
This adds support for passing through a context to use for rpc requests in a backwards compatible fashion. This is similar to the way it was added to the standard go `database/sql` library, so it will be familiar to Go developers.

### DIFF
--- a/src/oca/go/src/goca/acl.go
+++ b/src/oca/go/src/goca/acl.go
@@ -17,6 +17,7 @@
 package goca
 
 import (
+	"context"
 	"encoding/xml"
 	"fmt"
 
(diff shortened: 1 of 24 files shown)


treywelsh (approved):
I'm not an OpenNebula team member but I contribute to Goca sometimes so I add this unofficial approval as a small hint to the maintainers.
As said in https://github.com/OpenNebula/one/issues/5688#issuecomment-1033614684 it's a breaking change for peoples that already implement RPCCaller interface (one more method is required to be defined now) but it's for users that define their own client, doesn't seems a big deal to me.

You could potentially improve your PR in adding a comment for each new methods

paczerny:
Hi @peterwillis ,
we are going to merge this PR soon. We have new commit for goca in master. Before merge I'll do following steps:
* rebase this branch to master and force-push
* Add MarketPlace::EnableContext method
* Add simple test for non-trivial context - users can use it as example how to use the context

paczerny (approved):

paczerny:
Already merged https://github.com/OpenNebula/one/commit/9bc5fb82ff44398915408590e8100a0a3dbfaf2e

PR #5782, opened 2022-03-24
F #5785: fix escape string in golang client template

jokimina:


rsmontero:
Merged as per 995f2d137373564b4f58886e7ec725de3c47548a

THANKS

Issue #5869, opened 2022-06-08
vCPU models incompatible with the host CPU reported by the monitoring (KVM)


atodorov-storpool:
**Description**
KVM hosts with AMD EIPYC processors are becoming popular these days. But there are different CPU model emulation capabilities.
The current monitoring probes are reporting all possible CPU models. Thes leads to an option to select a model vCPU that is not supported by the host CPU.

**To Reproduce**
Try to instantiate a VM with _CPU_MODEL=SandyBridge-IBRS_ on AMD EPYC-IBPB powered host.
The instance will fail because of missing spectre mitigation emulation in the EPYC CPU.

> Wed Jun  8 14:01:12 2022 [Z0][VMM][I]: error: the CPU is incompatible with host CPU: Host CPU does not provide required features: spec-ctrl

**Expected behavior**
The monitoring probe should report only the usable by the host CPU models.

**Details**
 - Affected Component: [Monitoring]
 - Hypervisor: [KVM]
 - Version: [5.12+]

**Additional context**
The im/kvm-probesd.d/host/system/machines_models.rb script is querying `virsh cpu-models`
```
virsh -r -c qemu:///system cpu-models x86_64
486
pentium
pentium2
pentium3
pentiumpro
coreduo
n270
core2duo
qemu32
kvm32
cpu64-rhel5
cpu64-rhel6
kvm64
qemu64
Conroe
Penryn
Nehalem
Nehalem-IBRS
Westmere
Westmere-IBRS
SandyBridge
SandyBridge-IBRS
IvyBridge
IvyBridge-IBRS
...
```
But according to `virsh domcapabilities kvm` we have:
```xml
  <cpu>
    <mode name='host-passthrough' supported='yes'/>
    <mode name='host-model' supported='yes'>
      <model fallback='forbid'>EPYC-IBPB</model>
      <vendor>AMD</vendor>
     .....
    </mode>
    <mode name='custom' supported='yes'>
      <model usable='yes'>qemu64</model>
      <model usable='yes'>qemu32</model>
      <model usable='no'>phenom</model>
      <model usable='yes'>pentium3</model>
      <model usable='yes'>pentium2</model>
      <model usable='yes'>pentium</model>
      <model usable='no'>n270</model>
      <model usable='yes'>kvm64</model>
      <model usable='yes'>kvm32</model>
      <model usable='yes'>cpu64-rhel6</model>
      <model usable='no'>coreduo</model>
      <model usable='no'>core2duo</model>
[...]

PR #5870, opened 2022-06-08
B #5869: Filter out unsupported CPU models

atodorov-storpool:
Decided to format it as a patch than an
entire rewrite of the probe script.
This could be rewritten to populate the
models array with the supported only.


christian7007 (approved):

rsmontero:
Thanks!

Merged as per 7961ae2416814fda0bacf93f5fc52cdc0e779643

PR #5941, opened 2022-08-09
B-5914: fix double lines

treywelsh:
PR #5932 introduced a bug that duplicate some template lines. It's due to the two WriteString calls, only one call was needed


rsmontero:
merged as per 24a3c4cf20f20dc7f1a8df33a70351c5e7b2bb4f

Issue #5968, opened 2022-09-08
nVidia GPUs not listed when configured for PCI-Passthrough


gavin-cudo:
**Description**
nVidia GPUs are not listed under a PCI devices on a host configured with PCI-Passthrough.

**To Reproduce**
Configure a host with nVidia GPUs for PCI-Passthrough as per the documentation at https://docs.opennebula.io/6.4/open_cluster_deployment/kvm_node/pci_passthrough.html

Set the filter under /var/lib/one/remotes/etc/im/kvm-probes.d/pci.conf on the frontend to be:

```
:filter:
  - '*:*'
:short_address: []
:device_name: []
```

**Expected behavior**
All PCI devices are listed under a `onehost show <host_id>` including nVidia GPUs.

**Actual behavior**
All PCI devices are listed except nVidia GPUs.

**Details**
 - Hypervisor: KVM
 - Version: 6.4.0 CE and 6.4.0 Enterprise

**Additional context**
GPUs are listed fine on the host with:

```
lspci -nn -d 10de:*
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
01:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
02:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
02:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
81:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
81:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
82:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
82:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
83:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
83:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
84:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
84:00.1 Audio device [0403]: NVIDIA Corporation Device [10de:228b] (rev a1)
c1:00.0 VGA compatible controller [0300]: NVIDIA Corporation Device [10de:24b0] (rev a1)
[...]

PR #5982, opened 2022-09-28
 B #5968: Prevent vGPU support breaking PCI passthrough on NVIDIA cards

JungleCatSW:
The new pci.rb was by default merging a the NVIDIA device vendor ID with what users configured in pci.conf

This means that vGPU could not be switched off and hosts that were set up for PCI passthrough of NVIDIA devices would not run VMs.

vGPU support has been commented out in pci.conf so that existing one deployments don't break on upgrade but if users need vGPU they can set it in pci.conf.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,6 +70,7 @@ share/esx-fw-vnc/*.rpm
 share/esx-fw-vnc/.vagrant*
 
 *.vscode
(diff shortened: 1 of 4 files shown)


rsmontero:
THANKS!!!!, several fixes are already upstream based on your contribution

PR #6011, opened 2022-11-07
use literal heredoc to prevent bash evaluation of stdin data

TheJJ:
when the stdin of a command contains bash-evaluated characters, execution will fail with funny errors like:

```
bash: line 2: : command not found
```

this patch prevents the bash evaluation through a quoted here-doc word.

`man 1 bash` says:
> If word is unquoted, all lines of the here-document are subjected to parameter expansion,
> command substitution, and arithmetic expansion, the character sequence \\&lt;newline&gt; is ignored,
> and \ must be used to quote the characters \, $, and `.

we did not check if this allows unintended code execution through the input data, e.g. the configuration xml.

there's a lot more `<< EOT` here-docs in the code base which are unquoted - probably they can have the same unforseen side-effects.

### DIFF
--- a/src/mad/ruby/OpenNebulaDriver.rb
+++ b/src/mad/ruby/OpenNebulaDriver.rb
@@ -108,7 +108,7 @@ class OpenNebulaDriver < ActionManager
                                          @timeout)
         elsif options[:ssh_stream]
             if options[:stdin]
-                cmdin = "cat << EOT | #{command}"
+                cmdin = "cat << 'EOT' | #{command}"
                 stdin = "#{options[:stdin]}\nEOT\n"
             else
                 cmdin = command


TheJJ:
Our error was triggered by the following XML snippet, no idea how it initially ended up in there.
Every attampt of migration caused another error, which again inhibits migration.
The <code>\`</code> triggers code execution, thus trying to run `'` as command...
```
[...]
<USER_TEMPLATE><ERROR><![CDATA[Thu Nov  3 13:15:39 2022: MIGRATE: bash: command substitution: line 2: unexpected EOF while looking for matching `'' 
bash: command substitution: line 3: syntax error: unexpected end of file 
bash: command substitution: line 2: unexpected EOF while looking for matching `'' 
bash: command substitution: line 3: syntax error: unexpected end of file
bash: command substitution: line 2: unexpected EOF while looking for matching `''
bash: command substitution: line 3: syntax error: unexpected end of file error: internal error: process exited while connecting to monitor: 2022-11-03T12:15:37.964718Z
qemu-kvm-one: -blockdev {"driver":"rbd","pool":"one_disks_ssd","image":"one-7-18-0","server":[{"host":"10.0.18.1","port":"6789"},{"host":"10.0.18.129","port":"6789"},{"host":"10.0.18.130","port":"6789"}],
"user":"one","auth-client-required":["cephx","none"],"key-secret":"libvirt-3-storage-auth-secret0",
"node-name":"libvirt-3-storage","cache":{"direct":true,"no-flush":false},"auto-read-only":true,"discard":"unmap"}:
error connecting: Connection timed out Could not migrate some-uuid to somehost.lol ExitCode: 1]]></ERROR>
<HOT_RESIZE><CPU_HOT_ADD_ENABLED><![CDATA[YES]]></CPU_HOT_ADD_ENABLED>
[...]
```

TheJJ:
did anyone have time to have a look already? :)

paczerny:
Yes, will be merged soon. The fix works
I just have some issues running full integration tests. I have to make sure it doesn't break anything. It will be included in next release.

TheJJ:
Did you also fix the other heredocs already?

paczerny:
@TheJJ Thank you for the contribution. The fix was merged both to EE and CE versions. https://github.com/OpenNebula/one/commit/beba07c57636a5bc034a279aba06914c57f492ad

The same fix was applied also to Virtual Network Driver. Other occurrences of heredoc in shell scripts should be fine as they don't read the stdin and they can't be quoted, because variable expansion is needed.

TheJJ:
great! thanks for opennebula, a very useful tool!

Issue #5681, opened 2022-01-06
Support for contextualization CDROM stored on a block device


atodorov-storpool:
**Description**
I think that the issue is related to https://github.com/OpenNebula/one/issues/3245 but about the contextualization CDROM which I believe could be considered as a "special" volatile disk.
We already have a workaround implemented([here](https://github.com/OpenNebula/addon-storpool/blob/master/vmm/kvm/attach_disk.storpool) and [here](https://github.com/OpenNebula/addon-storpool/blob/master/vmm/kvm/deploy-tweaks.d.example/volatile2dev.py)) that worked flawlessly until the release of qemu-kvm v6.0.0. Initially, we thought that the issue is a libvirt bug that on eject reverts the device type of the cdrom to _file_ but we were proven [wrong](https://gitlab.com/libvirt/libvirt/-/issues/264).

**Use case**
Store the contextualization CDROM to a (shared)block device.

**Interface Changes**
I think that idea shared in the public commit by @paczerny at https://github.com/OpenNebula/one/commit/ba2aef58d65d97a6c14aa992a92ef9eac5fc7115 should be extended to handle the contextualization CDROM too. Also, the patch(slightly modified version of [this](https://github.com/OpenNebula/addon-storpool/blob/master/patches/vmm/5.12/reconfigure.patch) one) that I'll attach to this issue is addressing the contextualization reconfiguration (we are currently hot-patching our customers with this patch that could be affected).

**Additional Context**
Besides StorPool, we believe that other addon storage drivers could benefit from this feature too.

## Progress Status
- [ ] Branch created
- [x] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

rsmontero:
Yes you are right

atodorov-storpool:
I think the implemented functionality is incomplete. I will try to write a patch in a bit.

PR #6041, opened 2022-12-13
F #5681: disk source element depends on disk type

atodorov-storpool:
following the domain XML definition https://libvirt.org/formatdomain.html
```
_source_
  Representation of the _disk source_ depends on the _disk type_ attribute value as follows:
  _file_
    The _file_ attribute specifies the fully-qualified path to the file holding the disk. Since 0.0.3
  _block_
    The _dev_ attribute specifies the fully-qualified path to the host device to serve as the disk. Since 0.0.3
```

Also, I took the liberty to remove the _Image::FILE_ case form the switch as it is the default one.


rsmontero (approved):

rsmontero:
THANKS :pray: merged as per c51c35f810d55633a8024dbad56dd7e1fe7e13e6

PR #6055, opened 2022-12-15
F #6054: GOCA - make AR representation more generic

### LINKED ISSUE #6054: [GOCA] make address range representation more generic
/!\ To report a **security issue** please follow this procedure:
   [https://github.com/OpenNebula/one/wiki/Vulnerability-Management-Process]

**Description**
I closed #6019 for this more generic one

There is several issues that require to update the address ranges structures in the terraform provider  (https://github.com/OpenNebula/terraform-provider-opennebula/issues/192 and https://github.com/OpenNebula/terraform-provider-opennebula/issues/376).
I realized the XML parsing structure lack of flexibility, we could add custom attributes in an address range.

To remains simple and be more consistent with the way we manage other "sub"-resources (for instance disk and NIC for VMs), I propose to remove the address range structure and use something fully dynamic instead.

It breaks for those that would already use the address range structure.

**Expected behavior**
Be able to parse all attributes of an adress range

**Details**
 - Affected Component: [e.g. Sunstone, Scheduler or Storage]
 - Hypervisor: [e.g. KVM]
 - Version: [e.g. 5.4.12 or development]

**Additional context**

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

treywelsh:
Make address range representation more generic, this allow to solve several issues in the provider and make goca a bit more consistent.

Close #6054

### DIFF
--- a/src/oca/go/src/goca/schemas/virtualnetwork/virtualnetwork.go
+++ b/src/oca/go/src/goca/schemas/virtualnetwork/virtualnetwork.go
@@ -55,45 +55,8 @@ type VirtualNetwork struct {
 	Template             Template            `xml:"TEMPLATE"`
 
 	// Variable parts between one.vnpool.info and one.vn.info
-	ARs  []AR         `xml:"AR_POOL>AR,omitempty"`
-	Lock *shared.Lock `xml:"LOCK,omitempty"`
-}
-
-type AR struct {
-	ID                string  `xml:"AR_ID,omitempty"`
-	GlobalPrefix      string  `xml:"GLOBAL_PREFIX,omitempty"` // minOccurs=0
-	IP                string  `xml:"IP"`                      // minOccurs=0
-	MAC               string  `xml:"MAC,omitempty"`
-	ParentNetworkARID string  `xml:"PARENT_NETWORK_AR_ID,omitempty"` // minOccurs=0
-	Size              int     `xml:"SIZE"`
-	Type              string  `xml:"TYPE"`
-	ULAPrefix         string  `xml:"ULA_PREFIX,omitempty"` // minOccurs=0
-	VNMAD             string  `xml:"VN_MAD,omitempty"`     // minOccurs=0
-	MACEnd            string  `xml:"MAC_END,omitempty"`
-	IPEnd             string  `xml:"IP_END,omitempty"`
-	IP6ULA            string  `xml:"IP6_ULA,omitempty"`
-	IP6ULAEnd         string  `xml:"IP6_ULA_END,omitempty"`
-	IP6Global         string  `xml:"IP6_GLOBAL,omitempty"`
-	IP6GlobalEnd      string  `xml:"IP6_GLOBAL_END,omitempty"`
-	IP6               string  `xml:"IP6,omitempty"`
-	IP6End            string  `xml:"IP6_END,omitempty"`
-	UsedLeases        string  `xml:"USED_LEASES,omitempty"`
-	Leases            []Lease `xml:"LEASES>LEASE,omitempty"`
-
-	// Not filled with Info
-	Allocated string `xml:"ALLOCATED"`
-}
-
-type Lease struct {
-	IP        string `xml:"IP,omitempty"`
-	IP6       string `xml:"IP6,omitempty"`
-	IP6Global string `xml:"IP6GLOBAL,omitempty"`
-	IP6Link   string `xml:"IP6LINK,omitempty"`
-	IP6ULA    string `xml:"IP6ULA,omitempty"`
-	MAC       string `xml:"MAC,omitempty"`
-	VM        int    `xml:"VM,omitempty"`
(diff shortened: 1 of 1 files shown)


rsmontero:
Sure thing! THANKS!!!!

sk4zuzu (commented):
Hello @treywelsh

This is of course a breaking change, code like this:
```go
	vn, err := controller.VirtualNetwork(vnID).Info(false)
	if err != nil {
		log.Fatal(err)
	}

	IP := vn.ARs[0].IP

	fmt.Println(IP)
```
will have to be changed to
```go
	vn, err := controller.VirtualNetwork(vnID).Info(false)
	if err != nil {
		log.Fatal(err)
	}

	IP, _ := vn.ARs[0].Get(keys.IP)

	fmt.Println(IP)
```

Also the terraform provider build fails now:
```shell
terraform-provider-opennebula$ go build
# github.com/OpenNebula/terraform-provider-opennebula/opennebula
opennebula/helpers_vnet.go:19:27: undefined: virtualnetwork.AR
opennebula/helpers_vnet.go:130:23: invalid operation: attachedAR.ID == arIDStr (mismatched types func() (int, error) and string)
opennebula/resource_opennebula_virtual_network.go:855:54: undefined: virtualnetwork.AR
opennebula/resource_opennebula_virtual_network.go:950:34: cannot use vn.ARs[0].ID (value of type func() (int, error)) as type string in argument to strconv.ParseInt
opennebula/resource_opennebula_virtual_network.go:960:40: vn.ARs[0].Size undefined (type virtualnetwork.AddressRange has no field or method Size)
opennebula/resource_opennebula_virtual_network.go:961:44: vn.ARs[0].IP undefined (type virtualnetwork.AddressRange has no field or method IP)
opennebula/resource_opennebula_virtual_network.go:1111:53: undefined: virtualnetwork.AR
opennebula/resource_opennebula_virtual_network_address_range.go:227:13: undefined: virtualnetwork.AR
opennebula/resource_opennebula_virtual_network_address_range.go:231:14: invalid operation: a.ID == d.Id() (mismatched types func() (int, error) and string)
```
Just a question, have you considered adding to the AR struct a `dyn.Vector` to store custom attributes there and keep the original struct (something similar to how `UserTemplate` is handled for VMs)? :thinking:

treywelsh:
Thank you for your answer !

Yes it's a breaking change, I'll take care of updating the provider dependency on Goca if this one is merged.
It would break for people that use the address range struct elements, in the case they update their Goca dependency (this would require a manual action), and in this case it wouldn't be painful to migrate.

In the past I tried to mix in a structure "static" fields with "dynamic" fields in Goca and finally for simplicity I dropped the idea.
As far as I remember with the `any` tag for the XML module in go it would be possible to do something like:
```
type AR struct {
    ID string `xml:"AR_ID,omitempty"`
    ...
    Dynamic TemplateAny `xml:",any"`
}
```
This would require to update this code (if it's possible to reuse code parts), add a struct like `type TemplateAny struct {...}` with it's own unmarshal method: https://github.com/OpenNebula/one/blob/master/src/oca/go/src/goca/dynamic/unmarshal.go

To be honest I don't really like this code part and I wouldn't be happy to make it more complex, but maybe that there is a better way to write this. I'm open to discuss.

sk4zuzu (approved):
Yes, I was expecting code like that and I don't like it much either. This PR for sure solves the "ipam" and the "custom name" problems in the simplest way. LGTM

rsmontero:
Merged as per 462e888259363b1fd4c590d77c5c7b561b340856

THANKS!

treywelsh:
I introduced regression here, we loosed the capability to retrieve the Lease section of the address range when unmarshalling the vector.
If the address range is a `Vector`, then he's not able to contains the leases section which is itself a vector.

It's easy to see running a piece of code like this:
```
package main

import (
	"fmt"
	"log"

	"github.com/OpenNebula/one/src/oca/go/src/goca"
)

func main() {

	client := goca.NewDefaultClient(
		goca.NewConfig("XXX", "XXX", "http://XXX/RPC2"),
	)
	controller := goca.NewController(client)

	vNetID := XXX

	vn, err := controller.VirtualNetwork(vNetID).Info(false)
	if err != nil {
		log.Fatalf("info call: %s", err)
	}

	for _, ar := range vn.ARs {
		leases, err := ar.Get("LEASES")
		if err != nil {
			fmt.Printf("failed to get leases\n")
			continue
		}
		fmt.Printf("leases: %+v", leases)
	}
}
```

I didn't test my PR enough and I was influenced by my previous contributions.

So, I'll propose a new PR in which I added a commit that reverts my last changes and propose a new solution that try to introduce minimal changes and to reuse code.
To reuse existing code I introduced a new type called `Pairs`, which is a subset of a `Vector`, then I use struct embedding to make content and methods from `Pairs` available to `Vector`.

Apologies to the ONE team

PR #6104, opened 2023-01-30
F-6105: introduce pairs

### LINKED ISSUE #6105: [GOCA] fix virtual network regression
/!\ To report a **security issue** please follow this procedure:
   [https://github.com/OpenNebula/one/wiki/Vulnerability-Management-Process]

**Description**

Initial issue: https://github.com/OpenNebula/one/issues/6054
Related PR: https://github.com/OpenNebula/one/pull/6055

The goal was to improve address range parsing to be able to manage custom tags.
However the solution I initially proposed in the PR above introduced a regression: the `LEASES` section is not parsed anymore.

**To Reproduce**
Fill this piece of code with the configuration and run: 
```
package main

import (
	"fmt"
	"log"

	"github.com/OpenNebula/one/src/oca/go/src/goca"
)

func main() {

	client := goca.NewDefaultClient(
		goca.NewConfig("XXX", "XXX", "http://XXX/RPC2"),
	)
	controller := goca.NewController(client)

	vNetID := XXX

	vn, err := controller.VirtualNetwork(vNetID).Info(false)
	if err != nil {
		log.Fatalf("info call: %s", err)
	}

	for _, ar := range vn.ARs {
		leases, err := ar.Get("LEASES")
		if err != nil {
			fmt.Printf("failed to get leases\n")
			continue
		}
		fmt.Printf("leases: %+v", leases)
	}
}
```

**Expected behavior**
We should be able to see the LEASES and the custom tags.

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

treywelsh:
I introduced a regression in PR #6055
This PR revert the changes I made and propose an alternative solution.

To reuse existing code I introduced a new type called `Pairs`, which is a subset of a `Vector`, then I use struct embedding to make content and methods from `Pairs` available to `Vector`.

Apologies to the OpenNebula team for the noise around this

Close #6105

### DIFF
--- a/src/oca/go/src/goca/dynamic/dyntemplate.go
+++ b/src/oca/go/src/goca/dynamic/dyntemplate.go
@@ -42,10 +42,12 @@ type Pair struct {
 	Value   string `xml:",innerxml"`
 }
 
+type Pairs []Pair
+
 // Vector contains an array of keyvalue pairs
 type Vector struct {
 	XMLName xml.Name
-	Pairs   []Pair
+	Pairs
 }
 
 type TemplateAny struct {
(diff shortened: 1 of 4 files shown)


sk4zuzu:
Hello @treywelsh :)

Can we do something about these tests (update or remove) -> https://github.com/treywelsh/one/blob/F-6054-introduce-pairs/src/oca/go/src/goca/dynamic/dyntemplate_test.go ? :thinking:

treywelsh:
Hi,

This is an old file that was never updated so I just updated it to have minimal testing

sk4zuzu (approved):
LGTM :+1:

PR #6157, opened 2023-03-20
Change `u_int16_t` type for `uint16_t`

brodriguez-opennebula:
The types `u_intXX_t` (underscore after the `u`) are not supported
  by newer versions of GCC and `uintXX_t` should be used instead.

  In the file `src/svncterm_server/genfont.c` there were still some
  variables using the old type `u_int16_t`. They have been fixed.


rsmontero:
Merged as per 8c5dcbc17ffcc71c4efd18b6e7221d726bd822be

Issue #3999, opened 2019-11-28
option to add elements via onedb change-body


atodorov-storpool:
**Description**
Currently onedb change-body could update and delete an XML element but cannot add a new one.

**Use case**
When there is a change that needs to be propagated to the VM's configuration. For example to add FILTER_*_SPOFING to a NIC

**Interface Changes**
onedb change body to add an element instead of update when the element is not found in the XML

**Additional Context**

## Progress Status
- [ ] Branch created 
- [ ] Code committed to development branch
- [ ] Testing - QA
- [ ] Documentation 
- [ ] Release notes - resolved issues, compatibility, known issues
- [ ] Code committed to upstream release/hotfix branches
- [ ] Documentation committed to upstream release/hotfix branches

al3xhh:
PRs to merge in 5.10 and master:

code: https://github.com/OpenNebula/one/pull/4036
docs: https://github.com/OpenNebula/docs/pull/865

al3xhh:
Thank you so much for your contribution @atodorov-storpool !!

I have changed a bit the code and used the opportunity to lint the Ruby code.

rsmontero:
Merged in master and one-5.10 branch for 5.10.2 release

feldsam:
Hello @atodorov-storpool , @rsmontero , how can I add 'ID' to list of 'IMAGES' in datastore? I tried this:

`onedb change-body datastore --id 109 /DATASTORE/IMAGES/ID 451 --append --dry`, but it changes IDs of all other images:

```
<DATASTORE>
  ...
  <IMAGES>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID>451</ID>
    <ID><![CDATA[451]]></ID>
  </IMAGES>
  ...
</DATASTORE>
```

Thank you for help!

atodorov-storpool:
Hi @feldsam, @rsmontero 

[...]

PR #6170, opened 2023-04-03
M #-: onedb change-live --append with multiindex

atodorov-storpool:
do not update the matching elements on append
following comment in #3999


feldsam:
tested and append working good now

dann1 (approved):

Issue #6099, opened 2023-01-27
Fault Tolerance hook script improvements


OpenNebulaSupport:
**Description**
Improve the handling of local datastore VMs with the Fault Tolerance hook script.

**Use case**
When using VMs on a hosts which have mixed shared and local datastores the only options right now are to have the virtual machines be recreated or deleted during a failure.  This would allow the script to migrate shared VMs while ignoring locally stored VMs.

**Additional Context**
The script will need to check all disks of a virtual machine to determine if it is shared or local, which will determine behavior. When the fault tolerance script is configured to migrate any VMs on the failed host, this should skip over any VMs which contain at least one locally stored disk while still migrating any VMs which are fully on shared storage.

## Progress Status
- [x] Code committed
- [x] Testing - QA
- [x] Documentation (Release notes - resolved issues, compatibility, known issues)

onenhansen:
PR for script: https://github.com/OpenNebula/one-ee/pull/2583
PR for docs: https://github.com/OpenNebula/docs/pull/2550

PR #6182, opened 2023-04-18
F #6099: Improve FT Hook Script host_error.rb

onenhansen:
Adds checks to the "migrate" option of the Fault Tolerance hook script `host_error.rb` to skip virtual machines which are hosted on the failed host but are not deployed to a shared datastore.

The whole datastore pool is gathered at first, and that pool is used to reference the shared value to keep the API calls low.

This resolves https://github.com/OpenNebula/one/issues/6099

Should apply to 6.6 and up

### DIFF
--- a/share/hooks/ft/host_error.rb
+++ b/share/hooks/ft/host_error.rb
@@ -122,7 +122,7 @@ end
 
 mode    = nil    # **must** be set to something other than nil using the options
 force   = false  # By default, don't recreate/delete suspended and poweroff VMs
-repeat  = 2      # By default, wait for 2 monitorization cycles
+repeat  = 2      # By default, wait for 2 monitorization cycles
 fencing = true
 
 opts = GetoptLong.new(
@@ -248,6 +248,13 @@ if OpenNebula.is_error?(rc)
     exit_error "Could not get vm pool"
 end
 
+ds_pool = OpenNebula::DatastorePool.new(client)
+rc = ds_pool.info_all
+
+if OpenNebula.is_error?(rc)
+    exit_error "Could not get datastore pool"
+end
+
 # STATE=3: ACTIVE (LCM unknown)
 # STATE=5: SUSPENDED
 # STATE=8: POWEROFF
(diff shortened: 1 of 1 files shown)


tinova (approved):

share/hooks/ft/host_error.rb
```diff
@@ -283,6 +290,15 @@ def states_xpath(*arr)
             log "delete #{vm_id}"
             vm.delete
         when :migrate
+            vm_ds_id  = vm.retrieve_elements("/VM/HISTORY_RECORDS/HISTORY[position()=last()]/DS_ID")[0]
```
rsmontero, line 293: linting " --> ' I think there is no interpolation so single quotes?

rsmontero: Also is this "exception" save? I mean retrieve elements what happen if no vm_sd_id is found?

share/hooks/ft/host_error.rb
```diff
@@ -283,6 +290,15 @@ def states_xpath(*arr)
...
+
+            ds_xpath  = "/DATASTORE_POOL/DATASTORE[ID=\"#{vm_ds_id}\"]/TEMPLATE/SHARED"
+            is_shared = ds_pool.retrieve_elements(ds_xpath)[0]
```
rsmontero, line 296: Same here what if no is_shared?

rsmontero: Maybe migrate or not migrate by default?

share/hooks/ft/host_error.rb
tinova: good catch (pun intended) :p

share/hooks/ft/host_error.rb
tinova: > Maybe migrate or not migrate by default?

the idea is to not migrate non shared VMs

share/hooks/ft/host_error.rb
```diff
@@ -283,6 +290,15 @@ def states_xpath(*arr)
...
+
+            if is_shared == "NO"
+                log "Skipping VM #{vm_id} deployed on non-shared datastore"
+                next
+            end
```
rsmontero, line 301: So suggestion could be a rescue for this block that would skip the migration

share/hooks/ft/host_error.rb
onenhansen: It should be safe, as in the vm_id array should only contain VMs which were either in states ACTIVE, SUSPENDED, or POWEROFF (3, 5, 8) when pulled and should therefore have been scheduled at least once.   I'll add a check for it in the event some edge case occurs that I'm unaware of.

share/hooks/ft/host_error.rb
onenhansen: This should also be safe, however if a datastore somehow doesn't contain SHARED it should skip and it seems this code would allow it to pass.  I'll add a check here as well.

share/hooks/ft/host_error.rb
onenhansen: After looking at the code again, I agree the rescue seems like the cleanest way about this

PR #6256, opened 2023-06-30
Allow migration ov LVM-backed VMs

EBCEEB:
Update fs_lvm_ssh/mv to allow VMs migration between hosts.


rsmontero:
It seems that if the LUNs are shared (which is an what the driver assumes) this code is not needed. We prefer to keep this requirement and not merge this logic in the drivers

PR #6381, opened 2023-11-02
Fix iptables commands on versions > 1.8.7

feldsam:
iptables version > 1.8.7 (nf_tables) on EL9 doesn't support --wait option.

https://manpages.debian.org/testing/iptables/ebtables-nft-save.8.en.html#DIFFERENCES_TO_LEGACY_IPTABLES

https://lwn.net/Articles/759184/


rsmontero:
Merged as per 2e27704a1df818a918c2597679eb4711c9daed3a

THANKS!

Issue #6382, opened 2023-11-03
broken local snapshotrevert action in one_vmm_exec


atodorov-storpool:
**Description**
local override of the snapshotrevert action is not working.

**To Reproduce**
* define a local override script for snapshotrevert in the VM_MAD ARGUMENTS `-l snapshotrevert=snapshot_revert.local ...`
* issue VM snapshot revert command

The remote snapshot revert command is called.

**Expected behavior**
The defined local script should be called

**Details**
 - Affected Component: [one_vmm_exec]
 - Hypervisor: [KVM]
 - Version: [6.8.0]

**Additional context**
The bug is introduced with https://github.com/OpenNebula/one/commit/2efd9766453a45aae69bcc5bbdf952c50c942d9c

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

edit: fix action name: it is revert not restore

atodorov-storpool:
I have a workaround, defining `... -l snapshot_revert=local_snapshot_revert ...` but it is inconsistent with the rest of the overrides (`snapshocreate` and `snapshotdelete`)

PR #6383, opened 2023-11-03
B #6382: Fix action name when snapshot_revert

atodorov-storpool:


rsmontero:
Indeed, this was preventing to run custom action scripts, merged as per 73e9bc721ed2e2946ef38e83b823a52e21ca63a3

PR #6457, opened 2024-01-12
Update onevcenter

Shanobi:
change first 'import' example from `onevcenter list` to 'onevcenter import`

### Description

Reading through the help menu, noticed that for the import examples, the first one showed a "list" command argument rather than import, I assume it should be 'import'
Also, noticed that the 'import_desc' is used for both `import` and `import_defaults`, not sure if this is the desired case

- [ ] master
- [ ] one-X.X

<hr>


tinova (approved):

tinova:
thanks!

Issue #6472, opened 2024-01-23
libfs.sh:remove_off_hosts not filtering hosts


atodorov-storpool:
**Description**
The `remove_off_hosts()` function is used by  `get_destination_host()` to get the live hosts from a given list (datastore's BRIDGE_LIST variable). But it returns the entire list always.

**To Reproduce**
* Have an OpenNebula with at least two hosts (host1, host2)
* Mark one of the hosts as offline (let's powe-roff host1)
* on a shell, `source /var/lib/one/remotes/datastore/libfs.sh`
* call remove_off_hosts "host1 host2"
The function will return both hosts

**Expected behavior**
The function should return only the online host (host2)

**Details**
 - Affected Component: [DS_MAD]
 - Hypervisor: [KVM]
 - Version: [6.x, others?]

**Additional context**
Reviewing the function there is a flaw, or maybe a changed behavior of the `onehost` filtering argument?
it looks like `onehost list --filter="STAT=off,STAT=err,STAT=dsbl"` is doing logical _AND_ to the filtered arguments, and as the hosts could not be simultaneously in `off`, `err`, and `dsbl` it is returning no hosts...

Inverting the logic to do exclusive filtering looks like solving the issue. 

## Progress Status
- [x] Code committed
- [x] Testing - QA
- [x] Documentation (Release notes - resolved issues, compatibility, known issues)

PR #6473, opened 2024-01-23
B #6472: Fix remove_off_hosts()

atodorov-storpool:
Return only the hosts in ON or UPDATE state (ONLINE or MONITORING_UPDATE). The original logic include the INIT state too, but there is a possibility that a host could become in ERROR state on failed INIT.

- [ ] master
- [ ] one-X.X

<hr>


MarioRobres:
Hi @atodorov-storpool! Thanks for your input on the ticket and this PR. We've pursued a similar approach to yours, but with some adjustments. Our goal was to address this bug while preserving the original function's logic. However, your contribution shed light on an important issue with how different options were handled in the filtering process. We appreciate your contribution.

Issue #6689, opened 2024-08-09
Could not reconfigure a VM with context disk type=block


atodorov-storpool:
**Description**
This is another perspective of issue #6688. Please feel free to merge it although it is regarding another component of of opennebula (the VMM_MAD for KVM).

When a the contextualization CDROM is a block device an extra argument `--block` should be added to the `change-media` command used by the `vmm/reconfigre` script.

**To Reproduce**
* deploy a VM with context cdrom of type block
* try to reconfigure the VM (update configuration).

```
Fri Aug 9 10:32:55 2024 [Z0][VMM][I]: Command execution fail (exit code: 1): cat << 'EOT' | /var/tmp/one/vmm/kvm/reconfigure '73b45748-1863-4d19-95ab-42144817190a' 'hda' '/var/lib/one//datastores/0/0/disk.1' 0 u24onekvm2
Fri Aug 9 10:32:55 2024 [Z0][VMM][E]: reconfigure: Command "virsh --connect qemu:///system change-media 73b45748-1863-4d19-95ab-42144817190a hda /var/lib/one//datastores/0/0/disk.1 --insert" failed: error: Failed to complete action insert on media
Fri Aug 9 10:32:55 2024 [Z0][VMM][I]: error: internal error: unable to execute QEMU command 'blockdev-add': 'file' driver requires '/var/lib/one//datastores/0/0/disk.1' to be a regular file
Fri Aug 9 10:32:55 2024 [Z0][VMM][I]: Could not insert CDROM /var/lib/one//datastores/0/0/disk.1 to hda
Fri Aug 9 10:32:55 2024 [Z0][VMM][I]: ExitCode: 1
Fri Aug 9 10:32:55 2024 [Z0][VMM][I]: Failed to execute virtualization driver operation: reconfigure.
Fri Aug 9 10:32:55 2024 [Z0][VMM][E]: UPDATECONF: ERROR: reconfigure: Command "virsh --connect qemu:///system change-media 73b45748-1863-4d19-95ab-42144817190a hda /var/lib/one//datastores/0/0/disk.1 --insert" failed: error: Failed to complete action insert on media error: internal error: unable to execute QEMU command 'blockdev-add': 'file' driver requires '/var/lib/one//datastores/0/0/disk.1' to be a regular file Could not insert CDROM /var/lib/one//datastores/0/0/disk.1 to hda ExitCode: 1
```

**Expected behavior**
VM reconfigure should succeed.

**Details**
[...]

PR #6690, opened 2024-08-09
B #6689: append --block to change-media call in reconfigure

atodorov-storpool:
### Description

when the $ISO_PATH is a block device.
So libvirt's change media should be called with additional argument `-block`

- [x] master
- [x] one-6.10

<hr>


1gramos (approved):

rsmontero:
THANKS! Merged as per d5b5a4d5a42729d7eb2895739a6b91cf7e821205

This will be also included in 6.10.1 release

Issue #6710, opened 2024-09-05
[FSunstone] Keep me logged in option


atodorov-storpool:
**Description**
FSunstone login do not honor `keep_me_logged(_in)` server configuration

**To Reproduce**
The login dialog, keep me logged in checkbox.

**Expected behavior**
should disappear when configured in _/etc/one/fireedge/sunstone/sunstone-server.conf_
**Details**
 - Affected Component: [FSunstone]
 - Hypervisor: [KVM]
 - Version: [6.10]

**Additional context**
The configuration variable differ from the one looked up in the code...

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

PR #6711, opened 2024-09-05
B #6710: (FSunstone) keep_me_logged variable fix

atodorov-storpool:
### Description

Sync variable name with RSunstone and FSunstone's code

- [X] master
- [X] one-6.10

<hr>


vichansson (approved):
LGTM!

tinova:
We'll make sure to add this into master, thanks for the heads up @antalvrom !

Unfortunately we cannot take this to one-6.10 since we don't change conf files in maintenance releases, but we'll add a note in the documentation

PR #6737, opened 2024-09-26
Added serial attribute for disks in VM

jfeflinski:
### Description

Sometimes I need to have unique serials for disk in the VM - serials are created from the VM ID and next number, e.g.:

```
7771-1
7771-2
7771-3
7771-4
.
.
.
7771-N
```

- [ ] master
- [ ] one-X.X

<hr>

### DIFF
--- a/src/vmm/LibVirtDriverKVM.cc
+++ b/src/vmm/LibVirtDriverKVM.cc
@@ -546,6 +546,7 @@ int LibVirtDriver::deployment_description_kvm(
     string discard;
     string source;
     string clone;
+    string serial;
     string blk_queues;
     string shareable;
     string ceph_host;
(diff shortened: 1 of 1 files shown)


rsmontero:
Self-note before merging, from libvirt documentation:

```
If present, this specify serial number of virtual hard drive. For example, it may look like <serial>WD-WMAP9A966149</serial>. Not supported for scsi-block devices, that is those using disk type 'block' using device 'lun' on bus 'scsi'. Since 0.7.1
```

* Format, length is probably not a problem
* Type, we need to check disk is not of type block

rsmontero:
We merged a re-designed implementation of this PR for 7.0, Thanks for your contribution

jfeflinski:
Thanks @rsmontero!

PR #6762, opened 2024-10-23
KVM: Monitor VM IP address using qemu-guest-agent

jimis:


src/im_mad/remotes/kvm-probes.d/guestagent.conf
```diff
@@ -4,9 +4,18 @@
...
 :commands:
   :vm_qemu_ping: "one-$vm_id '{\"execute\":\"guest-ping\"}' --timeout 5"
+  # The first non-loopback address
+  :vm_ip_address:       >
+                        one-$vm_id '{"execute":"guest-network-get-interfaces"}' --timeout 5 |
+                        jq '{"return" : [ .return[]."ip-addresses"[]|select(."ip-address-type"=="ipv4" and (."ip-address"|startswith("127.")|not))."ip-address" ][0]}'
+  # Array of all IP addresses
+  :vm_ip_address_list:  >
+                        one-$vm_id '{"execute":"guest-network-get-interfaces"}' --timeout 5 |
+                        jq '{"return" : [ .return[]."ip-addresses"[]."ip-address" ]}'
```
jimis, line 21: This command does not work currently because OpenNebula tries to parse the JSON response as VM_TEMPLATE syntax. It should work after #6765 is fixed.

dann1:
Hi @jimis. Thanks a lot for the extremely detailed feedback.

- We have updated the monitoring code to address all of the problems you mentioned.
- The configuration files were also updated.
  - Your two commands were added with a small change. We changed the jqs a bit since the parsing failed when some VMs had some network interfaces configured in a certain way. For example, docker0.
  - The guest_info shown on the example docs was also added
  - only the ping is enabled by default. The rest are commented out.

The fixes will be coming on 6.10.2 and the master branch has been updated as well. Please check out [the changes](https://github.com/OpenNebula/one/commit/fcd8eb06e525cbed966332781c211fd61b6e5386).

PR #6805, opened 2024-11-29
M #-: Append ACK to README

pedroielpi3:
### Description

Added ACKs.

- [ ] master
- [ ] one-X.X

<hr>


pedroielpi3:
Need to check option for decreasing the font size before issuing the PR

pedroielpi3:
Will not use different font size, reopening PR

PR #7072, opened 2025-06-02
Fix for getting a clean struct in each VM looped in scheduler

tsarajar:
### Description

When looping through VMs in the scheduler, not all previous data was cleared before or at get_capacity. This leaked old VM requirements to the next one blocking the scheduler in certain circumstances. This fix will ensure a clean struct for each VM in the loop.

- [ ] master
- [ ] one-X.X

<hr>

### DIFF
--- a/src/scheduler/src/sched/Scheduler.cc
+++ b/src/scheduler/src/sched/Scheduler.cc
@@ -873,8 +873,6 @@ void Scheduler::match_schedule()
 {
     VirtualMachineXML * vm;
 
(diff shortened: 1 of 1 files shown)


paczerny (approved):

paczerny:
It's already fixed in master (version 7.0). Apply only to 6.10 branch

rsmontero:
OK THANKS! :heart: 

Merged in 6.10 branch for 6.10.5 release

Issue #7432, opened 2026-01-06
deploy failure when  802.1q driver is used but PHYDEV element in the template is an empty string


atodorov-storpool:
**Description**
On VM deploy there is an error logged
```
Mon Jan  5 11:24:33 2026 [Z0][VMM][I]: Command execution fail (exit code: 1): cat << 'EOT' | /var/tmp/one/vnm/802.1Q/pre
Mon Jan  5 11:24:33 2026 [Z0][VMM][E]: pre: Command "sudo -n ip link add link name .255 mtu 1500 type vlan id 255" failed.
Mon Jan  5 11:24:33 2026 [Z0][VMM][E]: pre: Cannot find device "name"
Mon Jan  5 11:24:33 2026 [Z0][VMM][I]: Cannot find device "name"
Mon Jan  5 11:24:33 2026 [Z0][VMM][I]: ExitCode: 1
Mon Jan  5 11:24:33 2026 [Z0][VMM][I]: Failed to execute network driver operation: pre.
Mon Jan  5 11:24:33 2026 [Z0][VMM][E]: DEPLOY: 802.1Q: ERROR: pre: Command "sudo -n ip link add link name .255 mtu 1500 type vlan id 255" failed. ERROR: pre: Cannot find device "name" Cannot find device "name" ExitCode: 1
Mon Jan  5 11:24:33 2026 [Z0][VM][I]: New state is POWEROFF
```
**To Reproduce**
configure the bridges manually on the host
Define a VM with VNIC with 802.1q driver but leave phydev an empty string
The vm will fail (with the above message)

**Expected behavior**
The vm shoild start following the coded logic when there is no phydev element defined.

**Details**
 - Affected Component: [Networking]
 - Hypervisor: [KVM]
 - Version: [6.10, 7.0]

**Additional context**
A patch that extends the checks of `...nil?` with `...empty?` is on the way...

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

rsmontero:
Thanks! merged as per 31cdddf00ae667d6126b0b1aeae56b7b6cd3d70b

PR #7433, opened 2026-01-06
B #7432: test PHYDEV for empty string

atodorov-storpool:
### Description

The change extends the tests for nil(not defined) with a test for an empty string to cover the case when the element is defined but has no value.

- [x] master
- [x] one-6.10
- [x] one-7.0

<hr>


paczerny:
Merged as https://github.com/OpenNebula/one/commit/31cdddf00ae667d6126b0b1aeae56b7b6cd3d70b

Issue #7586, opened 2026-04-08
queues attribute in disk driver element is only supported for virtio bus


atodorov-storpool:
**Description**
The `DEFAULT_ATTACH_VIRTIO_BLK_QUEUES` variable configured in the `kvmrc` file is applied when attaching a SCSI disk to a VM. Effectively failing the attach operation.

**To Reproduce**
1. Set DEFAULT_ATTACH_VIRTIO_BLK_QUEUES in the kvmrc file and update the hosts
2. Download a VM from the Marketplace
3. Instantiate the VM without altering the VM configuration
4. Attach a VM disk

The attach procedure will fail with the following error:
`Wed Apr  8 13:40:46 2026 [Z0][VMM][E]: ATTACHDISK: virsh --connect qemu:///system attach-device 094f4b75-5323-4a91-a43f-3ea60f7c7c05 <(cat <<EOT<disk type='block' device='disk'><driver name='qemu' type='raw' cache='none' io='native' discard='unmap' queues='4'/><source dev='/var/lib/one//datastores/0/1/disk.2' ></source><target dev='sdb'/><address type='drive' controller='0' bus='0' target='1' unit='0'/></disk>EOT)" failed: error: Failed to attach device from /dev/fd/63error: unsupported configuration: queues attribute in disk driver element is only supported for virtio busCould not attach /var/lib/one//datastores/0/1/disk.2 (sdb) to 094f4b75-5323-4a91-a43f-3ea60f7c7c05
`
**Expected behavior**
The disk attach should succeed.

**Details**
 - Affected Component: [Virtualization]
 - Hypervisor: [KVM]
 - Version: [7.0, 7.2]

**Additional context**
Another issue is clear in the data of the reported error message above.

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

rsmontero:
THANKS

Merge as per 60d54ced03fbeb6c207b2eb132ec2132dd756e94

PR #7587, opened 2026-04-08
B #7586: on attach apply default virtio-blk queues to virtio disks only

atodorov-storpool:
### Description

Add a test for the known TARGET prefix for the virtio-blk disks

- [x] master
- [x] one-7.4
- [x] one-7.0

<hr>


rsmontero:
THANKS merged as per 60d54ced03fbeb6c207b2eb132ec2132dd756e94

Issue #7588, opened 2026-04-08
default IOTHREAD no applied when attaching a virtio disk


atodorov-storpool:
**Description**
When attaching a virtio disk to a VM, the iothread is missing.

**To Reproduce**
1. Configure IOTHREADS = "1" in the vmm_exec_kvm.conf.
2. Restart the opennebula service
3. Instantiate a VM
4. Attach a virtio disk to the VM. The disk is attached with:
`<driver name="qemu" type="raw" cache="none" io="native" discard="unmap" queues="4"/>`

**Expected behavior**
The disk should have an iothread.

**Details**
 - Affected Component: [Virtualization]
 - Hypervisor: [KVM]
 - Version: [7.0, 7.2]

**Additional context**
After restarting the VM via OpenNebula, there is an iothread defined in the disk definition:
`<driver name="qemu" type="raw" cache="none" io="native" discard="unmap" iothread="1" queues="4"/>
`
I think "to have something rather than nothing" is better in this case. It is clear that replicating the round-robin logic across multiple places is hard, but a default value for a given dataplane inb the kvmrc file could be a good start.

Also, referring to my very, very old opinion that the right place for building the VM definition XML is in the VMM driver, not in the core of OpenNebula. Then the round-robin logic would be in one function used in both cases - on VM definition and on disk hotplug...

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

rsmontero:
THANKS

Merge as per 60d54ced03fbeb6c207b2eb132ec2132dd756e94

PR #7593, opened 2026-04-09
B #7588: apply iothread on virtio-blk disk hotplug

atodorov-storpool:
### Description

The proposed patch handles a typo in the VM metadata parser. 
Also, when the target is a `vd`(virtio-blk) disk:
 * Check if there are iothreads defined for the domain.
 * If there is speciffic thread id defined for the disk
 * If no - the optional default attach virtio iothread id
 * check the values: use the max iothreads if the suggested iothread is bigger than the defined iothreads, or do not define if the suggested iothread is 0 or less.

- [x] master
- [x] one-7.0
- [x] one-7.2

<hr>


rsmontero:
THANKS merged as per 60d54ced03fbeb6c207b2eb132ec2132dd756e94

Issue #7675, opened 2026-05-13
option for host-specific tproxy configuration


atodorov-storpool:
**Description**
The change is to extend the tproxy service with an option to merge a host-specific configuration with the global one. Usually, only the `remote_addr` field needs to be altered so the given host's tproxy service to know the specific service IP to communicate.

**Use case**
When there are separate clusters defined in OpenNebula for end-customer/rack/firewall-zone separation so there are different datastores, hosts and networks it is easier to place a per-host configuration instead of doing some routing magics to route the service traffic to a single IP address on the controller

**Interface Changes**
The change is triggered only when a given configuration file is found on the hosts (/var/lib/one/tproxy_host.config). Otherwise there is no change in the tproxy behavior.

**Additional Context**

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

atodorov-storpool:
Re-thinking a bit, I reached the conclusion that as we need to alter only the IP address, something like the `vnc-local` for the listen address in the VNC configuration...
So the provided patch is to handle the when a (resolvable) name is provided for `:remote_addr`

PR #7677, opened 2026-05-13
F #7675: option to configure FQDN for remote_addr

atodorov-storpool:
### Description
This way the resolver on the host will provide the IP address for tproxy to communicate with the opennebula-gate service. So, with a DNS view or a line in /etc/hosts file on the host(s) the tproxy service will pick the desired IP to connect to the opennebula-gate endpoint.

- [x] master
- [x] one-7.2

<hr>

### DIFF
--- a/src/vnm_mad/remotes/lib/tproxy.rb
+++ b/src/vnm_mad/remotes/lib/tproxy.rb
@@ -44,7 +44,20 @@ module VNMMAD
                      && (nets & nic.slice(:network, :network_id).values.map(&:to_s)).empty?
 
                 next if conf[:service_port].nil?
-                next if conf[:remote_addr].nil? || conf[:remote_addr] !~ Resolv::IPv4::Regex
+                next if conf[:remote_addr].nil? || conf[:remote_addr].to_s.strip.empty?
+
+                unless conf[:remote_addr] =~ Resolv::IPv4::Regex
+                    begin
+                        remote_ip = Resolv.getaddress(conf[:remote_addr])
+                        next unless remote_ip =~ Resolv::IPv4::Regex
+
+                        OpenNebula::DriverLogger.log_info "Resolved remote_addr #{conf[:remote_addr]} to #{remote_ip}"
+                        conf[:remote_addr] = remote_ip
+                    rescue Resolv::ResolvError
+                        OpenNebula::DriverLogger.log_warning "Can't resolve remote_addr #{conf[:remote_addr]}. Skipping."
+                        next
+                    end
+                end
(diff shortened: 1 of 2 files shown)


src/vnm_mad/remotes/lib/tproxy.rb
```diff
@@ -44,7 +44,20 @@ def self.setup_tproxy(nic, direction)
...
                 next if conf[:service_port].nil?
-                next if conf[:remote_addr].nil? || conf[:remote_addr] !~ Resolv::IPv4::Regex
+                next if conf[:remote_addr].nil? || conf[:remote_addr].to_s.strip.empty?
+
+                unless conf[:remote_addr] =~ Resolv::IPv4::Regex
+                    begin
+                        remote_ip = Resolv.getaddress(conf[:remote_addr])
+                        next unless remote_ip =~ Resolv::IPv4::Regex
+
+                        OpenNebula.log_info "Resolved remote_addr #{conf[:remote_addr]} to #{remote_ip}"
```
sk4zuzu: Shouldn't `OpenNebula::DriverLogger.log_info` be used here? 🤔

src/vnm_mad/remotes/lib/tproxy.rb
```diff
@@ -44,7 +44,20 @@ def self.setup_tproxy(nic, direction)
...
+                        conf[:remote_addr] = remote_ip
+                    rescue Resolv::ResolvError
+                        OpenNebula.log_warning "Can't resolve remote_addr #{conf[:remote_addr]}. Skipping."
```
sk4zuzu: Shouldn't `OpenNebula::DriverLogger.log_warning` be used here? 🤔

sk4zuzu:
Maybe indicating inside OpenNebulaNetwork.conf (in comments) that using FQDN is now a possibility would help some users. :thinking:

src/vnm_mad/remotes/lib/tproxy.rb
atodorov-storpool: yep, forgot that the logging was refactored somewhere between 6.x and 7.x
will force-push updated patch

src/vnm_mad/remotes/lib/tproxy.rb
atodorov-storpool: yep, forgot that the logging was refactored somewhere between 6.x and 7.x
will force-push updated patch

sk4zuzu (approved):
LGTM

rsmontero:
THANKS! merged as per af0b2a4f7d

Issue #7662, opened 2026-05-08
Improve the transactional nature of VM migrations


balazsbme:
**Description**
The current migration procedure lacks transactionality (either everything gets migrated, or nothing). When a migration fails due to destination issues (e.g., a permission error), the system metadata can improperly update to show the VM on the target host despite the data never being transferred. 

If an operator then improperly attempts to recover the VM using the "recover with success" option (ignoring GUI warnings, and documentation regarding when this option should be used, see [here](https://docs.opennebula.io/7.0/product/operation_references/opennebula_services_configuration/troubleshooting/#:~:text=the%20following%20options%3A-,%2D%2Dsuccess,-%3A%20If%20the%20operation) ), the VM remains in an inconsistent state. Attempting to migrate the VM back to the original host at this stage triggers a cleanup process that deletes the partial folder on the original host—destroying the only existing copy of the VM's disk and resulting in permanent data loss.

**To Reproduce**
1. Create a VM and power it off (assume it is on `host1`).
2. Inject a permission error on another host (`host2`): `chown -R root:root /var/lib/one/datastores/0/`
3. Try to migrate the VM from `host1` to `host2`.
4. The migration fails due to the permission issue. Observe in GUI/CLI that the VM appears to be running on `host2`, even though it does not have any data there.
5. Fix the permission issue on `host2`: `chown -R oneadmin:oneadmin /var/lib/one/datastores/0/`
6. Simulate an operator mistake by using the “recover --success” option.
7. Try to start the VM that appears to be on `host2` now. It fails (`disk.0` not found, as it was never transferred).
8. Try to migrate back from `host2` to `host1`. The procedure cleans up the partial folder on `host1` (deleting the only valid copy of `disk.0`, causing data loss), and reports a successful migration.
9. The VM cannot start on `host1` with the same error: `disk.0` not found (at this point it is deleted permanently).

See VM logs attached here as an example of this reproduction: [reproduced-in-microenv-vm-logs.txt](https://github.com/user-attachments/files/27523448/reproduced-in-microenv-vm-logs.txt)

**Expected behavior**
Migration operations should be strictly transactional. If a migration fails during the transfer phase, the metadata should not falsely report the VM as residing on the target host (mitigating Step 4). Furthermore, the cleanup procedures during migration should have safeguards to prevent the deletion of the source disks if they are the only surviving copy, even in the event of an operator misusing the `recover --success` override.

**Details**
 - Affected Component: Core / Migration Lifecycle
[...]

PR #7682, opened 2026-05-15
F #7662: preserve rollback path for recover success

ktfh-claw:
## Summary
- stop treating `recover --success` on `PROLOG_MIGRATE_*_FAILURE` states as a direct boot/deploy success path
- route those failure states through the existing prolog-failure recovery path instead
- preserve explicit retry behavior while avoiding a synthetic success transition on the destination side

## Rationale
`PROLOG_MIGRATE_*_FAILURE` already means the transfer-manager prolog migration failed.
Calling `recover --success` from those states should not skip straight into boot/deploy as if the destination-side prolog had succeeded. The existing `trigger_prolog_failure()` path already performs the rollback/history handling for these migration failure states, so this change reuses that logic.

## Validation
- `git diff --check`
- successful `scons mysql=yes grpc=yes systemd=yes rubygems=yes -j32` build on Ubuntu build VM

## Notes
I did not replace the running disposable 7.2.0 lab with this branch because the patch was built from current `master` (`7.3.80.pre`), and forcing a master build over that package-based lab would create unnecessary version skew during validation.

### DIFF
--- a/src/lcm/LifeCycleActions.cc
+++ b/src/lcm/LifeCycleActions.cc
@@ -1358,18 +1358,14 @@ void LifeCycleManager::recover(VirtualMachine * vm, bool success,
         //----------------------------------------------------------------------
         case VirtualMachine::PROLOG:
         case VirtualMachine::PROLOG_MIGRATE:
-        case VirtualMachine::PROLOG_MIGRATE_FAILURE:
         case VirtualMachine::PROLOG_RESUME:
         case VirtualMachine::PROLOG_RESUME_FAILURE:
         case VirtualMachine::PROLOG_UNDEPLOY:
         case VirtualMachine::PROLOG_UNDEPLOY_FAILURE:
         case VirtualMachine::PROLOG_FAILURE:
         case VirtualMachine::PROLOG_MIGRATE_POWEROFF:
-        case VirtualMachine::PROLOG_MIGRATE_POWEROFF_FAILURE:
         case VirtualMachine::PROLOG_MIGRATE_SUSPEND:
-        case VirtualMachine::PROLOG_MIGRATE_SUSPEND_FAILURE:
         case VirtualMachine::PROLOG_MIGRATE_UNKNOWN:
-        case VirtualMachine::PROLOG_MIGRATE_UNKNOWN_FAILURE:
             if (success)
             {
                 trigger_prolog_success(vim);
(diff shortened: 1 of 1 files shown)


ktfh-claw:
Fresh clean-lab validation result for this patch:

I reran the migration/recovery test in a new isolated two-hypervisor disposable runtime instead of relying only on the earlier lab.

Environment:
- frontend/runtime: `10.1.1.130`, `ONE_LOCATION=/home/claw/one-issue7662-fresh130`, OpenNebula `7.3.80`
- host `0`: `10.1.1.132`
- host `1`: `10.1.1.133`
- test VM: `ID 0`, `vm7662-run`
- base image: CirrOS qcow2 imported as image `cirros7662`

Controlled repro in the fresh lab:
1. Booted the VM on host `0`.
2. Forced a target-side TM failure on host `1` with:
   - `chmod 000 /home/claw/one-issue7662-fresh130/var/datastores/0`
3. Ran:
   - `onevm migrate 0 1 --poff-hard`
4. Reproduced the expected migration failure in `var/vms/0/vm.log`:
   - `Error executing image transfer script: Error copying disk directory to target host`
   - `rm: cannot remove '/home/claw/one-issue7662-fresh130/var/datastores/0/0': Permission denied`
   - `tar: /home/claw/one-issue7662-fresh130/var/datastores/0: Cannot open: Permission denied`
   - `tar: Error is not recoverable: exiting now`
5. VM entered `PROLOG_MIGRATE_FAILURE` on host `1`.

Patched behavior check:
1. Restored target permissions:
   - `chmod 775 /home/claw/one-issue7662-fresh130/var/datastores/0`
2. Ran:
   - `onevm recover 0 --success`
3. Observed the important behavior again:
   - active host rewound back to source host `10.1.1.132`
   - history became:
     - `SEQ 0` -> `10.1.1.132`
     - `SEQ 1` -> `10.1.1.133`
     - `SEQ 2` -> `10.1.1.132`
   - recovery path transitioned through `PROLOG_MIGRATE` -> `BOOT_MIGRATE` on the source side, instead of treating the failed destination-side prolog migration as a direct success path
4. The restore then failed in this disposable lab because there was no checkpoint file, leaving:
   - `BOOT_MIGRATE_FAILURE`
   - `RESTORE: error: failed to get domain 'one-0'`
   - `Failed to open file '/home/claw/one-issue7662-fresh130/var/datastores/0/0/checkpoint': No such file or directory`

So the important assertion held again in a fresh clean two-host runtime: `recover --success` from `PROLOG_MIGRATE_FAILURE` rewinds to the source-host rollback path instead of pretending the destination-side prolog migration succeeded.

Issue #7691, opened 2026-05-18
opennebula-gate via tproxy broken because the network namespaces are dropping cross-subnet ARP requests


atodorov-storpool:
**Description**
VMs are unable to resolve the MAC address for local endpoints `169.254.16.9/24` because their ARP requests are silently dropped inside the network namespace by the kernel.

When the VNM driver creates a new network namespace, the interfaces inside inherit the host's default sysctl ARP policies. If the host OS is configured with strict ARP settings (when the default `arp_ignore` is greater than 0), the namespace enforces this strictness. Consequently, the namespace interface ignores and refuses to reply to any cross-subnet ARP requests originating from the VM.

**To Reproduce**
* Follow the installation guide and configure the tproxy for opennebula-gate service
* check that the arp_ingnore sysctl is enabled (it is not 0)
* Instantiate a VM with enabled opennebula gate
* try to ping 169.254.16.9 or issue request via the onegate tool (both will fail with network is unreachable.
VMs fail to connect to metadata services or route traffic properly because the necessary ARP replies are never generated by the namespace.

**Expected behavior**
The network namespace should reliably reply to local ARP requests from the VM, regardless of the host's underlying anti-ARP-flux or strict routing configurations.

**Details**
 - Affected Component: [Networking]
 - Hypervisor: [KVM]
 - Version: [6.10.1+]

**Additional context**

When the default libvirt network bridge virtbr0 is used for testing, there is another issue caused by the implemented SNAT (MASQUERADING). Aa trivial iptables rule on the nat table leverage the problem but I think this sub-issue is irrelevant as it is on another OSI layer and manifests only when the current issue is resolved.

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

PR #7692, opened 2026-05-18
B #7691: loosen inherited ARP strictness in netns

atodorov-storpool:
### Description
to allow cross-subnet ARP replies.

Network namespaces can inherit strict ARP policies (where arp_ignore is greater than 0) from the host's default configurations. When this strictness is inherited, it breaks cross-subnet ARP resolution. Specifically, the namespace silently drops and fails to reply to the VM's cross-subnet ARP requests for local endpoints, such as the 169.254.16.9.

This PR explicitly loosens the inherited ARP strictness by forcing arp_ignore=0 inside the namespace. By applying this to both all and the specific interface, it guarantees the namespace will successfully reply to the VM's ARP requests for local IPs.

- [x] master
- [x] one-6.10.1+
- [x] one-7.2

<hr>


sk4zuzu (approved):
LGTM. This is indeed precise fix. I've included it also here https://github.com/sk4zuzu/tproxy-rs/commit/1e1b9e3d26184f13604b48e6e58e2d3ef3bee690 in the integration tests of my rust version of tproxy. 🙏😇

atodorov-storpool:
@sk4zuzu
Took a look at the rust patch. Please note that when there is a dot ('.') in the interface name (think for vlan interfaces, for example `onebr0.1234`), the dot in the interface name should be replaced with '/' for the sysctl because the dot is reserved for path separation.

sk4zuzu:
@atodorov-storpool yes I noticed, and it's also correct, but the patch is just for integration tests it's unlikely I'll use interfaces with dots there. Thanks! :hugs:

rsmontero:
THANKS merged as per 0cebd42c8f

Issue #7915, opened 2026-07-16
one_drs ignores PREDICTIVE variable


atodorov-storpool:
**Description**
The `PREDICTIVE` setting in `one_drs.conf` (and the optional cluster `ONE_DRS` / `PREDICTIVE` attribute) is documented to blend current and forecasted resource usage into OPTIMIZE decisions. A value such as `0.3` should mean 30% forecast / 70% current.

In practice, predictive blending never takes effect: `_apply_predictive_adjustment()` always sees a predictive factor of `0`, so OPTIMIZE continues to use current usage only even when `PREDICTIVE` is set and forecast metrics are present.

**To Reproduce**
1. Enable monitoring forecasts so hosts/VMs expose `*_forecast` metrics.
2. Set `PREDICTIVE: 0.3` in `/etc/one/schedulers/one_drs.conf` (or in the cluster `ONE_DRS` template).
3. Trigger an OPTIMIZE plan and inspect how CPU/memory (etc.) usage is computed for the ILP.
4. Observe that only current usage is used (equivalent to `PREDICTIVE: 0`).

Forecasts are never blended in, regardless of the configured `PREDICTIVE` value.

**Expected behavior**
With `PREDICTIVE > 0` and non-zero forecasts, OPTIMIZE should blend current and forecast values for:

- VM CPU / net / disk usage
- host memory / CPU usage

when building the ILP inputs.

**Details**
 - Affected Component: [Scheduler]
 - Hypervisor: [KVM]
 - Version: [7.x]

**Additional context**

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

rsmontero:
THANKS Fixed as per a553e769a3fcabf0793eaa0abf6c47209c3f4481

PR #7917, opened 2026-07-16
B #7915: Fix PREDICTIVE config lookup in one_drs

atodorov-storpool:
### Description
Read the lowercase "predictive" config key while the config stores PREDICTIVE, so the predictive factor was always 0 and forecasts were never blended in. Also guard against a None value coming from a cluster ONE_DRS template without PREDICTIVE defined.

- [x] master
- [ ] one-X.X

<hr>


rsmontero:
merged in a553e769a3fcabf0793eaa0abf6c47209c3f4481

Issue #7962, opened 2026-08-07
[OneBex] incremental backups broken for thin LVM


atodorov-storpool:
**Description**
Making incremental backup on disks with more that 1(one) dirty extent are failing

**To Reproduce**
1. Define LVM-thin datastores
1. Instantiate a VM on them
1. Start a job inside the VM that writes at different offsets (copy/delete files, whatever)
1. Create first backup
1. Create sequential incremental backup

**Expected behavior**
The backups should be created

**Details**
 - Affected Component: [OneBEX]
 - Hypervisor: [KVM]
 - Version: [7.4]

**Additional context**
Extents are built with only `:start`/`:length` keys ([`lvm.rb:268-269`](https://github.com/OpenNebula/one/blob/090bafc6acbca9452138a3c0a0e203c234e6e34f/src/onebex/exporters/lvm.rb#L267-L273), [also](https://github.com/OpenNebula/one/blob/090bafc6acbca9452138a3c0a0e203c234e6e34f/src/onebex/exporters/lvm.rb#L111-L112), but the [merge logic](https://github.com/OpenNebula/one/blob/090bafc6acbca9452138a3c0a0e203c234e6e34f/src/onebex/exporters/lvm.rb#L279-L296) reads `extent[:end]` / `last[:end]`.

With 0 or 1 extents the loop body never runs, so it happens to work; with ≥2
changed extents it raises `ArgumentError: comparison of Integer with nil
failed`, `GET /images/:id/extents` returns 500, and the incremental backup of
any lvm-thin disk fails.

## Progress Status
- [ ] Code committed
- [ ] Testing - QA
- [ ] Documentation (Release notes - resolved issues, compatibility, known issues)

PR #7963, opened 2026-08-07
B #7962: fix normalize_extents in lvm OneBEX exporter

atodorov-storpool:
### Description

Use defined/populated elements

- [x] master
- [x] one-7.4

<hr>


rsmontero:
THANKS @atodorov-storpool  good catch :) Already merged as per 98ad38a05870cd3e856cf430892003c0a3eb12eb

